PR #595, opened 2022-07-12
Festo update.

Mobiuslau:
A script was written to add charts and songs to the corresponsing json collections for Jubeat. Song titles and artists were manually fetched. The `chartID` strings were generated using `secrets.token_hex(20)`, and were not checked for duplicates. New additions were appended to the end of the array, with `songID` starting from `1051`. The 6 charts per song were added in difficulty order `ADV`, `BSC`, `EXT`, `HARD ADV`, `HARD BSC`, `HARD EXT`, per song. The `displayVersion` strings are derived from the first digit in the `id`. The `versions` list contains `festo` for every chart which has appeared in Jubeat Festo before December 2020; it does not assume current state. Charts removed during the running of Festo have still received the `festo` tag.

### DIFF
--- a/database-seeds/collections/charts-jubeat.json
+++ b/database-seeds/collections/charts-jubeat.json
@@ -101654,7 +101654,8 @@
 		"songID": 901,
 		"tierlistInfo": {},
 		"versions": [
-			"clan"
+			"clan",
+			"festo"
 		]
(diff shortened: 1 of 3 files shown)


zkldi:
lgtm

PR #609, opened 2022-07-23
feat(#608): Add stardust and starlight tables

### LINKED ISSUE #608: Stardust/Starlight BMS tables support
These are two modern tables for the normal scale, which shouldn't matter too much for most people playing on insane but I think it's pretty worth it for newer players.

Stardust: https://mqppppp.neocities.org/ChartView.html
Starlight: https://djkuroakari.github.io/starlighttable.html

zkldi:
Fixes #608 

Self-explanatory. Also refactors where bms_tables are declared from.

We should probably add sieglinde values for these tables, if they're good?


zkldi:
Sieglinde handling of this can be done in the future, I guess.

PR #616, opened 2022-08-02
fix(docs): link properly points to tng-dev repo

j1nxie:
okay, this should *properly* fix everything. sorry for the trouble. 

don't code at 7 am after pulling an all nighter.


zkldi:
the docs are comically out of date and i need to get around to tidying them all up properly.

this will likely happen alongside the short guide.

zkldi:
anyway, i pushed some stuff to fix some more things wrong with the doc urls while i was there.

PR #641, opened 2022-08-21
Update old USC converts levels to the 1-20 scale

### LINKED ISSUE #607: SDVX1-2 Converts for USC have incorrect levels on the site
Amusing note: The checksums used by USC include the chart level; this means that people playing the game cannot change their levels locally, or they will no longer have the same chart checksum as everyone else. Wonderful!

However, we can change this in the seeds and it will propogate fine in the UI (and fix some faulty VF calcing)

Adamaq01:
Fixes #607 

There might still be some charts that were not updated but I think this covers a majority of the charts tracked by tachi


cg505:
LFG

zkldi:
lgtm

PR #642, opened 2022-08-22
Sieglinde V1

zkldi:
This PR adds support for Sieglinde V1, alongside all the scripts and magic to calculate it.

This adds Sieglinde Data to the following tables, applied in the following order:

```ts
[
	"Normal2",
	"Insane2",
	"Overjoy",
	"Satellite",
	"Stella",
	"Normal",
	"Insane",
]
```

*****

Other than that, this should sync with production pretty fine, and we'll just do a full site recalc.


zkldi:
That test fail is uh. bizarre, to say the least. Appears to be race-condition related somehow. Going to skip it as it's not pertinent to this stuff.

Issue #658, opened 2022-08-27
Organising Festo Final Charts


Mobiuslau:
Throughout its lifespan, Festo has added and removed charts such that some charts in the Festo folders are inaccessible for Final players. As such, a `festo final` tag should be added to the `versions` list for charts released in Festo Final. Moreover, it would be benificial to add Festo Final folders according to this tag.

zkldi:
This, or we just remove `festo` tags from songs that were in festo but aren't anymore.

Tachi convention is that `$version` is the state of the game as of its final update, so we should really just remove the `festo` tag from anything that was removed from `festo`. If this affects people on older datecodes, so be it. It will eventually resolve itself as people move forward.

Mobiuslau:
The latter is probably the most consistent solution. Having seperate Festo tables would probably be redundant for most players anyway.

The `festo` tags will be readjusted soon in that case.

zkldi:
Was fixed in #660

PR #660, opened 2022-08-28
Jubeat Versions List Assumes Final State for Festo.

Mobiuslau:
Fixes issue #658.

`festo` tag was removed from `versions` list if the chart was not present in Festo Final.


zkldi (commented):
lgtm

PR #661, opened 2022-08-31
Remove instanceof from score importing

zkldi:
Due to some unknown issues, `instanceof` appears to be randomly and spontaneously failing to match classes. There is almost no consistency to why this error is occuring, but it is annoyingly unstable and resulting in errors not being handled correctly. Given that we want errors to be handled correctly, this is a problem.


zkldi:
ok git, wyd.

Issue #633, opened 2022-08-18
BPI Chart says "You" instead of the username of the user you're viewing


zkldi:
UI bug, not amazingly significant but not too hard to patch I don't think?

zkldi:
fixed by removing the bpi chart (lol)

PR #663, opened 2022-09-03
Refactor Client to use GPT/UGPTContexts

zkldi:
This refactor enables a UGPT/GPT available "UGPTContext" which allows any component in those routes to hook into information about the logged-in user and the user's page we're viewing.

This is obviously convenient for a bunch of reasons, and helps us tackle things like #633.

Sadly, I'm not able to properly test this on my laptop as I don't have access to a set of testing data. We should *really* look into getting a bootstrappable local-test-db for this stuff.


zkldi:
Since this PR might not work, but I have no way of testing it locally, I'm going to push it to staging and test it more properly there.

Maybe things would suck a bit less if my PC with my SSH keys wasn't in storage :)

PR #666, opened 2022-09-03
[backport] add dive with ü

cg505:
backport #665


zkldi:
not sure how i feel about backports being merged like this, but it does make sense.

I use cherry picking at the moment because merging doesn't work (staging and release have unrelated histories)

PR #667, opened 2022-09-04
feat(server): have assets on localdev

zkldi:
These are lazily duplicated from `tachi-assets`; we keep them here so that the client looks less bad.


zkldi:
whoops this branch was made off of release rather than staging

it's trying to resolve disparate histories, so, i don't care.

PR #670, opened 2022-09-04
Sieglinde Tables

### LINKED ISSUE #643: Sieglinde Tables in-game
Make actual tables for the sieglinde things so people can use them in game a-la walkure suggestions. I guess.

hewoicvewse:
fixes #643

### DIFF
--- a/common/src/utils/util.ts
+++ b/common/src/utils/util.ts
@@ -253,3 +253,13 @@ export function FormatSieglindePMS(sgl: number): string {
 
 	return `●${(sgl - 12).toFixed(2)}`;
 }
+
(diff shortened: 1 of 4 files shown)


zkldi:
very happy with this PR. Adds a couple of hacks to create "GPT Specific" routes, which is a bit painful. Going to write some documentation for this quirk and the endpoints in general.

Other than that, this seems great to me. We could probably improve performance by caching the output of this, but it's not really significant.

Also, I broke it by moving the imports around. Going to look into not breaking that as much.

zkldi:
lgtm

zkldi:
backported to release

PR #671, opened 2022-09-05
add exscore to sdvx eamuse csv imports

cg505:
This is not ideal at the moment because score dedup will prevent this from taking effect for any existing PBs, but it's better than nothing.


zkldi:
backported, but we're having deploy turbulence atm

PR #673, opened 2022-09-08
add support for WaccaMyPageScraper CSV import

cg505:
This allows people to upload data from https://github.com/XezolesS/WaccaMyPageScraper, which is a widely used tool to save data from the WACCA webui (known as MyPage) before it goes offline at the end of September.
Before it goes offline, we prefer the current userscript system. However, if people join Kamaitachi after MyPage goes offline, that won't be available. But they may have saved their profile data using this other project, so it will be nice to support it for the future.


server/src/lib/score-import/import-types/file/wacca-mypage-scraper/types.ts
```diff
@@ -0,0 +1,31 @@
+export interface MyPageRecordsRawCsvRecord {
+
```
cg505, line 2: please tell me how to make this pass lint

zkldi: it's a cadence bug. ignore it with // eslint-ignore

zkldi:
status check on this?

cg505:
got busy but this is still on my radar. mostly just pending tests.

zkldi (approved):

zkldi:
feels real weird having 0 open prs now...

PR #675, opened 2022-09-12
Modify database-seeds/repo.ts to not use cd

Bottersnike:
Rather than chaining a `cd` command within `asyncExec`, we can instead utilise the `cwd` option within `child_process.exec`. This is slightly cleaner, and is also the magic sauce to make `repo.ts` run on windows (I'm so sorry). This PR splits up all the conjoined commands within `repo.ts` and instead lets typescript handle sequential execution of commands.


zkldi:
turn cadence on

Bottersnike:
yes daddy

zkldi:
Should be good now. Will wait for CI to pass and then merge from bed.

PR #700, opened 2022-10-02
Permit empty "about me"s

Bottersnike:
Amends the react code to not pass null and adds a server test for this case (though the server was already accepting zero-length strings).

### DIFF
--- a/server/src/server/router/api/v1/users/_userID/router.test.ts
+++ b/server/src/server/router/api/v1/users/_userID/router.test.ts
@@ -120,6 +120,20 @@ t.test("PATCH /api/v1/users/:userID", async (t) => {
 		t.end();
 	});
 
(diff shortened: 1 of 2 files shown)


server/src/server/router/api/v1/users/_userID/router.test.ts
```diff
@@ -120,6 +120,20 @@ t.test("PATCH /api/v1/users/:userID", async (t) => {
 		t.end();
 	});
 
+	t.test("Shouldn't allow empty strings for about me.", async (t) => {
```
zkldi, line 123: drunk test name. `should allow`, not `shouldn't`

zkldi (changes requested):

zkldi (approved):

PR #713, opened 2022-10-13
Use mask-image for banner transparency

Bottersnike:
Wow! What an unexpected PR!

Rather than using a colour gradient, use a pseudo-element with `mask-image` applied to it, allowing the original background colour to show through.

### Notes:
- This explicitly bumps the `<td>` up to z-index 1. As many other UI elements make use of 0-5 for general layout purposes within metronic I do not believe this to cause any issues, but this has not been extensively confirmed. Menus and modals appear to use 90+.

### DIFF
--- a/client/src/_style/base.scss
+++ b/client/src/_style/base.scss
@@ -63,3 +63,24 @@
 .display-block {
 	display: block;
 }
+
+.banner-td {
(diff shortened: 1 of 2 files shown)


zkldi:
lgtm ty

PR #715, opened 2022-10-16
feat: baseline rivals support

zkldi:
This PR adds baseline rival UI support for Tachi. It does most of the stuff you'd expect from a normal rival implementation.

Still to go:

- rival feed
- pinning chart sets and doing stats on those
- snipe indicators
- seeing rival goals & quests


zkldi:
i don't like the errors its giving me so i'm merging anyway

PR #718, opened 2022-10-18
eAM CSV related changes

yahooeny:
Adds strange csv formatting for "Get Set, Go!", moves Praludium searchterms into altTitle

### DIFF
--- a/database-seeds/collections/songs-iidx.json
+++ b/database-seeds/collections/songs-iidx.json
@@ -12707,7 +12707,11 @@
 		"title": "VEGA"
 	},
 	{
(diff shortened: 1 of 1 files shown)


zkldi:
twenty three thousand changes?

yahooeny:
??? Weird.
Also forgot a comment somewhere so i quickly recommited

zkldi:
Merged by hand. Your git client is producing kinda broken diffs. I'm not sure how this has happened, but I would look into it :sweat:. 

Thanks for the PR! The changes have been applied.

Issue #690, opened 2022-09-29
Error Messages from imports are no longer displayed to the client


zkldi:
Consequence of bull now cleaning up after itself, the imports are deleted from redis, which also means the error messages are. We should store these permanently in MongoDB.

PR #726, opened 2022-10-19
Fix bug where all batch manual imports show as failed

Bottersnike:
Frontend types for `/imports/:importId` didn't match what the backend was sending, causing `undefined` propagation. This PR fixes the types, extracts the correct data we require from the API response, then also stores the full response details (the `songs`, `scores`, `sessions`, etc. that were missing from the types) in a new field `details`, which is nullable as this information is unique to `/imports/:importId` and is not present on other import related endpoints such as polling.


zkldi:
PR is correct. Fixes a different implementation bug with `useImport`. The more difficult #690 is a job-based issue where imports - upon failing - never save any information about their failing anywhere, so `poll-status` cannot retrieve anything about them.

PR #778, opened 2022-11-28
feat: make monorepo sane

ereti:
i need to apply this to everything else (only done common and server rn)


zkldi:
the monorepo man he imports everything like this

PR #827, opened 2022-12-28
feat: update mmdx seeds

j1nxie:
updating the seeds to match with live JP ratings.

currently a draft PR since i'm going through all the charts for each decimal level, shouldn't take too long to go through them all though as i only need to go down from 14.9 to 12.7.


zkldi:
:heart:

j1nxie:
short update: this hasn't fallen out of my radar, btw. i'm looking to get uni+ data and pull everything from there first, then update everything with festival changes.

PR #837, opened 2023-01-09
Tachi V3

zkldi:
This is a huge internal reworking of Tachi based on everything I've learned from v2. This fixes a bunch of major long-standing limitations with Tachi, and spends a lot of time simplifying all of the interfaces (especially for game/gpt configs).


zkldi:
Breaking changes so far:

- Classes are now stored as strings instead of integers. This includes changing all of the class processing code to work on records of strings rather than records of numbers
- `chart.rgcID` is gone. (was unused)
- `user.clan` is gone. (was unused)
- `chart.tierlistInfo` no longer exists, and is melded with `chart.data` on a game-by-game basis.
- Scores are no longer guaranteed to have `score`, `lamp`, `grade` or `percent`. Instead, the metrics for a score are declared in the GPT configuration.
- Enum Indexes (such as `score.scoreData.lampIndex`) are no longer stored like that. Instead they are stored in `.scoreData.enumIndexes`, so `.lampIndex` would be `.enumIndexes.lamp`, etc. This allows for any sort of enum index storing/querying.
- Goals no longer use `scoreData.score` etc. as keys, instead using `score`, `percent`, `lamp`, etc. (depending on what metrics are available).

Probably some other stuff I'm forgetting?

Currently around 200 tests are failing, so we'll just have to start analysing each failure. Long road.

zkldi:
we no longer meet 60% branch coverage so lets go do that i guess (lol)

zkldi:
Also it's worth noting that I also want to move `level` and `levelNum` to `chart.data` as they're gpt specific properties.

For example, some games only have numeric levels and the str equivalents are useless, and some games don't have interesting levels at all (BMS/PMS)

zkldi:
i've just gone and fucking made it all more complex but now that i've seen the light and power of this complexity i don't think i can go back

PR #838, opened 2023-01-10
Add festo-omni to versions in Jubeat seeds for omni charts.

Mobiuslau:
To support omni score submission, `festo-omni` was added to the `versions` list for each chart in omni.


zkldi:
no guarantees on this seeing merge/review for a while, sorry

zkldi:
all the raw data for me to work with is here but this branch needs to be kinda completely reworked to roll with the new seeds approach. if you've got the script you used to make this data, please run it again from a fresh copy of `staging`.

If you need any help dealing with the `git` side of things here, let me know in discord.

There are some significant changes to how charts are stored as of `2.4`, the most notable one being that `rgcID` and `tierlistInfo` no longer exist. For jubeat, this means you should no longer define those keys.

PR #865, opened 2023-02-15
Updated Infinitas entries for new monthlies + fix

justOkapi:
+3 (Sync, Nitrous Cannon; monthlies - 2 Beasts Unchained; incorrectly flagged)
-1 (デモーニッシュ; incorrectly flagged)


justOkapi:
IDK why it saw fit to believe I changed all the testing files.

zkldi:
windows uses a different way of going to the next line than linux. I'm not 100% on what the right way to get Git on windows to acknowledge this is, but it **should** convert windows style line endings (`CRLF`) to linux style (`LF`) by default.

Since it's clearly not (and you're not the only person this has happened to), i'll need to look into what the right setup option is.

zkldi:
Update: no, these files were just wrong in the repo. This PR is fine.

PR #875, opened 2023-03-07
Moved IsValidURL for Callback URLS to array based

Fluzzarn:
List taken from Twitter (lol)'s disallowed callback protocol list.

### DIFF
--- a/server/src/utils/misc.test.ts
+++ b/server/src/utils/misc.test.ts
@@ -15,6 +15,7 @@ t.test("#IsValidURL", (t) => {
 	t.ok(IsValidURL("http://example.com/suburl"));
 	t.ok(IsValidURL("http://example.com#href"));
 	t.ok(IsValidURL("http://example.com?querystring"));
+	t.ok(IsValidURL("tachi-fdsaf7324hf://example.com?querystring"));
 
 	t.notOk(IsValidURL("ftp://example.com"));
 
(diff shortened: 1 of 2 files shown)


zkldi:
what's this for? this array based solution requires active maintenance against future potentially insecure standards.

is there something wrong with just supporting HTTP and HTTPS? that's how we send the msgs anyway.

Fluzzarn:
The reason for this is because mobile apps don't use http/https for callback urls in an oauth2 flow. Both platforms have the app register a uri scheme that the OS knows to redirect to the app.

zkldi:
Interesting. Is there no way to make this an agreed upon constant so it can be whitelisted? otherwise, this blacklist would require constant maintenance and potentially be a security hole.

Fluzzarn:
Reading up on what [oauth recommends](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/) oauth recommends in this scenario and places usually go two routes:
1) Incorporate the app id into the url scheme
2) Have the url scheme follow a pattern that is validated on the server side e.g. tachiApp(whatever)://

The second one can work since it's still a whitelist for "http/https/whateverpattern" and doesn't require any UI changes

zkldi:
i like the pattern. is prefixing with `tachi-` ok?

zkldi:
not off my radar, just will be done later. probably thursday. remind me on discord if i haven't.

PR #879, opened 2023-03-15
Adjustments to searching

saintnoodle:
These changes included in this PR:

- Increase the chart search limit to 300 in both global search and /games/G/PT/songs (I don't have enough data to test whether this affects PBs, but it shouldn't since those limits are separate)
- Add "searchTerms" to searchFunctions sitewide to allow for filtering by searchTerms

The searchTerms filter will make more sense as I push changes to the database in the future


zkldi:
Why is the chart search limit now 300 instead of 100?

saintnoodle:
If you make a search that pulls more charts than 100, attempting to filter those results may result in an incomplete or empty search since you can only filter out of the results pulled from the main search. 300 was a number I found suitable to handle decently sized searches however 200-250 might be a good compromise if you’re concerned about performance.

Bottersnike:
fwiw from experience I've found "do excessive backend search then filter frontend" systems to end up really not working great, even if you start to crank the limit well past 300 or implement workarounds like searching in a loop until you have n results that match your frontend filters. Generally where possibly it's much nicer to have those filters all applied during database operations

zkldi:
figure out a way to implement it on the database then. mongodb doesn't support any sort of native searching that doesn't suck. our scale isn't large enough to warrant elasticsearch, and redissearch is too much of a pain to synchronise up with the state of the DB.

the filters aren't really that useful on the song search and are bad UX anyway, users want to just search for songs and the existence of two different inputs that filter the inputs in different ways (search vs filter) is just confusing in general.
i'm not entirely sold that the limit needs to be increased, as the search bar already listens to `searchTerms`.

PR #881, opened 2023-03-19
Adding support for jubeat music bar (+ seeds)

Meta-link:
This pull request adds the support of the music bar for jubeat scores. It contains score and seed data modifications, rendering of the music bar and updated seeds for every chart.

Possible improvements : 
- Vertically center the canvas in its div
- Find a solution to display the whole bar in the nested "play history" dropdown


zkldi:
change your git config please to use `pull.rebase` with `git config --global pull.rebase true`

PR #884, opened 2023-03-23
Fix the Notifications link in profile dropdown

Bottersnike:
Diff should be fairly obvious.

Food for thought: Why is this link even here? It's duplicating the function of the message icon that's perpetually on the header. Personal suggestion would be to remove that icon either always (and just show a `(1)` tag on the profile picture), or remove it when there are no unread notifications, cus it's a bit redundant and feels weird to have something like that that's barely used promoted hard by the UI.


zkldi:
i'll just remove this thing, it doesn't need to be here.

PR #895, opened 2023-04-09
Fixed missing Infinitas entries

justOkapi:
I have no clue how some of those slipped through the cracks, but here we are...


zkldi:
What are the mistakes you fixed?

zkldi:
I'm still not sure about this quest by the way, the goals just aren't particularly... interesting. Then again, INFINITAS players have brainworms, so idk.

justOkapi:
Midnight Datenshi was completely unflagged aside from the legacy Another chart, and somehow a bunch of recent additions were also not flagged.

As for the quest, it's not supposed to be interesting, it's completionist BS for brainwormed idiots like myself and is really more meant to make tracking such a stupid goal less tedious to input since the UI for adding goals leaves a lot to be desired when you're doing stuff in bulk, but I digest, if you don't want it on the site in an official capacity I have no problem with it not going up.

justOkapi:
Quest removed.  We can revisit this if people other than me clamor for it.

zkldi:
i have no strong opinions on the quest being here or not but we can revisit it at a later time maybe.

PR #896, opened 2023-04-12
enable barbatos novice imports 

saintnoodle:
0 = NOV yet is not parsed causing an "Invalid...Request" error.


zkldi:
excellent catch, thanks.

PR #898, opened 2023-04-20
feat: initial maimai support

beerpiss:
An attempt to reintroduce maimai "classic" support into Tachi.


zkldi:
is this ready for review? tests aren't passing.

common/src/constants/grade-boundaries.ts
```diff
@@ -129,12 +129,8 @@ export const MUSECA_GBOUNDARIES = MakeGradeBoundaries<GetEnumValue<"museca:Singl
 	傑G: 1_000_000,
 });
 
-export const MAIMAIDX_GBOUNDARIES = MakeGradeBoundaries<GetEnumValue<"maimaidx:Single", "grade">>({
-	D: 0,
-	C: 10,
-	B: 20,
-	BB: 40,
-	BBB: 60,
+const MAIMAI_COMMON_BOUNDARIES = {
```
zkldi, line 132: don't put this into a common thing, it adds indirection in the codebase for no reason. Just write it out twice; they're separate games.

server/src/game-implementations/games/maimai.ts
```diff
@@ -0,0 +1,233 @@
...
+//
+// There is also a dan bonus, up to 2.10 for Shinkaidan.
+//
+// This does not take into account the RECENT component.
+async function CalculateMaimaiRate(
+	game: Game,
+	playtype: Playtype,
+	userID: integer,
+) {
+	const user = await db["game-stats-snapshots"].findOne(
```
zkldi, line 64: This is the wrong place to fetch a users dan from. Use `game-stats` instead of `game-stats-snapshots`. The snapshots are purely for graphing purposes.

server/src/lib/migration/migrations/v3-props.ts
```diff
@@ -43,6 +43,7 @@ function perToScore(goal: GoalDocument) {
 		case "bms":
 		case "gitadora":
 		case "jubeat":
+		case "maimai":
```
zkldi, line 46: This migration script is for moving existing scores (those stored on tachi 2.3) to 2.4. This will never be ran again, so it needs no update.

zkldi:
left a review, also, please run the linter on your code, there are many minor formatting issues that can be automatically fixed.

server/src/game-implementations/games/maimai.ts
```diff
@@ -0,0 +1,233 @@
...
+					rate: { $max: "$calculatedData.rate" },
+				},
+			},
+			{
+				$sort: {
+					rate: -1,
+				},
+			},
+			{
+				$limit: 570,
```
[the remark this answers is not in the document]
beerpiss: That's how rating in maimai worked; an average of the top 570 songs is a part of the rating, with 570 being the number of songs available on FiNALE's release.

If this is too expensive, I suppose an average of the best 30 songs + dan bonus will suffice (or maybe ditch dan bonus entirely?)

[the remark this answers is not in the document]
zkldi: i think this metric is stupid then to be honest. We should just use `naiveRating` as we do for the rest of the games that have overly complex/stupid rating systems.

server/src/game-implementations/games/maimai.ts
beerpiss: marking as resolved since we moved entirely to naiveRating, no dans required.

server/src/lib/migration/migrations/v3-props.ts
beerpiss: the type checker isn't happy when i remove it

image

zkldi: ah, makes sense, even if it's nonsense and this migration has already ran.

saintnoodle:
There appears to be a whole block of songs between 800-851 that are just completely missing chart seeds

beerpiss:
> There appears to be a whole block of songs between 800-851 that are just completely missing chart seeds

Good catch - I've added the missing charts from FiNALE 1.99C

zkldi:
need more time to review this, sorry

zkldi:
at some point along the way i absolutely screwed this PR with some changes. going to hand-merge this now.

zkldi:
OK yeah this got bit really hard by a forcepush at some point such that it doesn't even think it's part of the repo anymore. This will take me some time to graft.

saintnoodle:
I had this issue on a couple of my branches fortunately I'm doing both of them from scratch so it doesn't matter.

PR #903, opened 2023-05-05
initial support for EXScore

saintnoodle:
Starting from SDVX EG Season 3, we now have a new max EXScore metric which allows us to store and calculate the difference between a the EXScore a player got against that chart's maximum EXScore. From this, we should be able to add and change the lamp to "S-PUC" when both values are exact. 

In this initial commit: 
- The fundamental support for max EXScore is added
- All charts now "know" their max EXScore
- PUC is changed to pink
- S-PUC is assigned gold
- The difference is displayed in the score cell on the client [e.g. 5133 EX (MAX -138)]
- The script for adding new seeds has been updated to retrieve the new max_exscore value from the XML and add it to seeds

In regards to the script, it may be useful to rename the old one and keep it around in case for whatever reason someone wants to use it on an pre EG Season 3 database.

All that's left from this point is to change the lamp to "S-PUC" when the EXScore of a play and the max EXScore of that chart match.


saintnoodle:
The lamp and score cells are now gold when exScore === maxExScore. Currently, the EXScore is set to display as a percentage when below 96% to mimic behaviour shown in game. We can either keep it this way or switch to using either a percentage or a difference. 

Further, EXScore isn't displayed at all if it isn't >= 1 since there's no point in showing the EXScore if it's 0. This is useful in cases where EXScore is grabbed from an import and converted from null to 0, like if a play was in light mode/skill analyser or from a session with S-Critical disabled. 

In regards to the seeds that just failed; those are songs that have been since removed from the game. We might need to make the max EXScore property completely optional to account for this to maintain backwards comparability with older versions that still have those songs.

Next, a method of sorting by the play EXScore / chart's max EXScore would be useful to allow for sorting by EXScore if those scores are at the max.

saintnoodle:
max exscore is now 0 for the handful of songs that don't have one. 

However it is **very important** to note that some of these charts are currently konaste exclusives and could be given maxExScore values at any time. These charts may need to be either manually adjusted or have the script that handles conversions from music_db.xml run a check to see if a given chart has maxExScore set to 0, and if it does, write the updated maxExScore value to the seeds.

The client handles a 0 max exscore by checking if maxExScore > 0 before allowing any EXScore element to render. S-PUC should also impossible based on this.

zkldi:
don't overload 0 to mean `null`. make the variable null and handle the case where it's null everywhere in the codebase.

saintnoodle:
We love typos. maxExScore is now null for any chart that does not have a value assigned in the database. Code to calculate and render the exscore has been moved to its own component and will only be used if it is given numbers. There are two functions based on whether a chart has a maxExScore or not. If there is no maxExScore and the score has an EXScore value, then a function is called to just display it as long as it's more than 0 to account for any import script bug that may occur or if for some reason a user intentionally submits an EXScore of 0.

PR #907, opened 2023-05-31
Added announced June Infinitas unlocks

justOkapi:
Holic in Infinitas is proof Konami loves me and wants me to be happy.

ヒマワリ is cool too I guess.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,4 +42,3 @@ failed-tests.log
 # bootstrap flag
 I_HAVE_BOOTSTRAPPED_OK
 database-seeds/scripts/rerunners/iidx/okapi-inf.js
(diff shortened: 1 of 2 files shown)


zkldi:
whats happening with the gitignore here? if this script is in `rerunners` it should be published in the repo.

justOkapi:
Literally the only thing that changes in my script is the songID, it's unnecessary for it to be updated every time I update the monthlies.

justOkapi:
And I *think* my script is already in the main staging branch.

justOkapi:
Confirmed my script is in the main staging branch already, added some commentary on how it works in case someone else wants to take a stab at it.

zkldi:
ah right makes sense. seems fair that this is untracked. simpler than making this script take command line args.

PR #913, opened 2023-07-23
update to bootstrap 5.3

saintnoodle:
## Updates
- Bootstrap to 5.3
- React Bootstrap to 2.9.0-beta.0
- Sass to 1.64.1

&nbsp;&nbsp;&nbsp;&nbsp; This update will also remove any metronic theme assets and attempt to fix any broken components as a result.
## The What and Why
&nbsp;&nbsp;&nbsp;&nbsp; The goal is to restore the site as close as possible to its original state after the upgrade and migration is complete. 
The motivation is to remove the metronic theme and the restrictions its proprietary license has on Tachi's source code so the client is easier to maintain and update. The update to Bootstrap 5 improves existing components like forms, adds the utility API, and allows for dark/light mode switching if wanted.
#### Further Ramblings Beyond
&nbsp;&nbsp;&nbsp;&nbsp; My further motivations are to allow for a better and complete implementation of the react-bootstrap library and to reduce the reliance on custom/inline css styles where possible. There are some cases where inline styles are acceptable however; 
- when one specific style is required but adding it to the global stylesheet makes no sense
- inserting styles dynamically from JS
- JS animations

Otherwise, Bootstrap already comes with a powerful utility class system and this should be the preferred method for styling custom components and pages as it encourages consistency and reduces complexity. It is already being used in a lot of areas, but there are some where it could be better utilised. Bootstrap 5 also opens up the utility API with Sass so we can expand and extend the utilities for what else we may need. All of this can be done gradually over time after this PR as part of updates, refactors, redesigns, and visual bug fixes.


saintnoodle:
I've caught as many visual bugs as I can though I don't have a full set of data to test with so there may be a few more I haven't seen. I will work on building a full set of testing data eventually.

I believe any more changes I want to make are beyond the scope of this PR so I'm comfortable to leave it here for now after it has been checked, tested and reviewed. Once I have the go ahead for merge, please allow me to remove any files I have made effectively redundant before the pull is made.

zkldi:
Is this PR actually ready for review or merge? there are loads of commits flying onto this branch.

saintnoodle:
I believe I've caught everything. I did a few last minute tests on my end and found a few more bugs and almost totally forgot about the login pages. Should be ready for review and preview now. I'm not going to touch this unless any issues are found.

zkldi:
I fixed some code that wasn't compiling, but I think it's likely that
the files were just unused, so it wasn't breaking the build.

-- other than that --

None of this affects the functionality of your PR, just some general
style tidbits:

- don't disable prettier. I know that the `${}` splitting is disgusting
  but that's life. prettier is there to brutally normalise all of the
  code for easier diffs.

- prefer `function` to `const foo = () => {}`. I don't think we use
  the anonymous assignment syntax for top level functions anywhere else
  in the project. This doesn't change anything really.

- Your /** */ JSDoc comments were missing proper indentation, so I added
  it.

- I removed some unused imports. These were likely left in unused by me,
  so i've removed them as they're definitely unimportant.

Everything else looks fine apart from the `useBreakpoint` function which
you mentioned had a fix elsewhere.

This looks great.

zkldi:
going to let this deploy on staging, give it an hour to simmer and let me poke around on it and then push fwd into main

PR #915, opened 2023-07-26
feat: update maimai dx seeds to festival plus

j1nxie:
charts for festival plus songs (except the 4 songs that are added in intl festival) still don't have correct chart constants just yet.


zkldi:
lgtm, thank you for this!

PR #918, opened 2023-07-30
Buncha Infinitas stuff

justOkapi:


justOkapi:
Thank [insert favorite higher power here] for the the DB manager

zkldi:
i'm the higher power here

zkldi:
i wonder why the bot doesn't re-comment when another subsequent change alters the seeds. it really should.

justOkapi:
> i wonder why the bot doesn't re-comment when another subsequent change alters the seeds. it really should.

it's your code my guy

PR #919, opened 2023-07-31
fix: landing page ghost button

saintnoodle:
#### Some bug fixes with a sprinkle of QoL improvements.

The home page for users logged out of Tachi would notice a blank button... It's now fixed

### Changes

The import method instructions have had alerts added to make warnings more noticeable and slightly easier to read (especially when the background is white....). Links have also had underlines attached to make them more noticeable.

The Profile Picture component is now unified, accepting a size prop and can also be rendered without a wrapping anchor.

User picture settings are now a bit clearer with the unset button more clearly stating what it will do with a red background. The deletion of your profile picture or banner now requires confirmation. Once an image has been loaded into the form, it can also now be reset. 

The reasoning for this ability to reset the form isn't primarily to provide the user with the ability to abort an upload that hasn't actually happened, but more so to reduce visual confusion with the buttons' functions and prevent someone from confusing the unset button *as* a cancel button, clicking it, and accidentally deleting their profile picture.

### Bug Fixes

- The "Your Profiles" dropdown would not update after a user imports scores for a game they haven't yet played until a hard refresh.
- In the user picture settings, auto margins were causing strange element spacing.

### DIFF
--- a/client/src/app/pages/dashboard/DashboardPage.tsx
+++ b/client/src/app/pages/dashboard/DashboardPage.tsx
@@ -20,11 +20,12 @@ import { UserContext } from "context/UserContext";
 import { UserSettingsContext } from "context/UserSettingsContext";
 import { ColourConfig, TachiConfig } from "lib/config";
 import React, { useContext, useMemo } from "react";
-import { Alert } from "react-bootstrap";
+import { Alert, Stack } from "react-bootstrap";
 import { Link, Route, Switch } from "react-router-dom";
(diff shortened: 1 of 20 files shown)


zkldi:
aftercare is an interesting word to use here

saintnoodle:
What would be more suitable

zkldi:
"fixes" would've been what i went for. aftercare has some interesting connotations lol

zkldi:
lgtm

PR #926, opened 2023-08-04
Outer Limits missing from Infinitas - fixed

justOkapi:


zkldi:
Very skeptical of this! [REMYWIKI](https://remywiki.com/OUTER_LIMITS) says that the charts in infinitas are SPN3, SPH6 .. but we have SPN5, SPH8 on record.

There's [OUTER LIMITS ALTERNATIVE](https://remywiki.com/OUTER_LIMITS_ALTERNATIVE) which has that data.

Those two charts are **often** mixed up in sources as they're sometimes the same song in older games, and iirc sometimes have the same title. It might be possible that the data for these two songs we have on record are incorrect from the get go.

Any chance you could investigate more into this?

justOkapi:
Yeah I'll look into it in a few minutes.  The one thing I can say beyond a shadow of a doubt is fervidex-static on my account is flagging the in game ID for Outer Limits as missing, and I confirmed it was not listed on Infinitas on Kamai.  I'll circle back with you in a bit.

PR #929, opened 2023-08-05
feat: sdvx max exscore

saintnoodle:
Add maxExScore to seeds and the script. charts with missing maxExScore is set to null. 

The only charts that should have no max exscore are charts that were removed before the season 3 update. if, for whatever impossible reason, these charts are given a max exscore, the script overwrites null if a value for that chart exists.

A method to display distance and hopefully sort by ExScore/maxExScore will follow in the future.


zkldi:
not accepting this if `maxExScore` is nullable. it shouldn't be nullable as the idea is nonsensical; all charts have a maxEXScore that can be figured out.

saintnoodle:
I'm not sure how we're supposed to calculate the maxExScore. For one, the charts that don't have a maxExScore are impossible to gain an ExScore for unless you're playing on an unsupported omnimix. These values are simply pulled straight from the game's database along with the other data that comes from that. We have the mathmatics, but I am not personally aware of a database that contains each chart's max count of chips, longs, and volts.

zkldi:
figuring it out is an implementation detail. the more important thing here is that the amount of possible EX on a chart is *not* an optional thing. it exists for all charts so making it nullable means that every bit of code that interfaces with it has to handle the case where the maxExScore isn't specified.

There's a similar parallel to draw with something like -- say -- if iidx's `notecount` field were nullable. That's not true at all, all charts have notes.

saintnoodle:
The amount of ExScore does not exist when playing Vivid Wave, which is the version the removed charts exist in. This is what I mean by it is impossible to gain an ExScore for these charts.

zkldi:
yes i know, but they still conceptually *have* an amount of ex score, which is some multiple of the notes in the chart. making this nullable means that all the code around ex score has to handle the case where exscore doesn't exist, but exscore *always* exists, because the charts always have a known amount of notes.

saintnoodle:
ExScore doesn't always exist since it's a metric that relies on S-Crits. S-Crits are only enabled on Valkrye cabinets (or VW nemsys cabinets modified to run in Valkrye mode) in normal, premium modes, and select battle modes. That is to say, in other modes like light mode or skill analyser, s-crits are always disabled. It's also an option to turn s-crits off in normal/premium mode. When s-crits are disabled the game exports "0" for both exscore and "just", the internal alias for s-crit.

Vivid Wave does export a "judge" field, and older games have "judge_checker" which does have 7 judgement points on chips even before they actually conceptualised s-crits, but in EG they managed to destroy the function that generates this data. This means the data from the judge metric is now gibberish. It does mean however that ExScores from previous games can be retroactively calculated, but how this data is collected and managed by networks, and subsequently our access to it, is unknown to me.

In regards to calculating max exscore for charts that were removed before they could be given one, the only way generating this data would be possible would be by having someone or something obtain a perfect score for all songs with the missing data and inspecting the packets the game sends to the server. The maximum value for each metric of the deleted charts isn't provided to us, and no one has collected it to my knowledge. The only other way is by observing footage or print-outs and manually counting...

All of this is to say that Konami decided it was a great idea to introduce a metric to the game and lock it down as a premium feature to more or less force arcades to buy the latest and greatest, very expensive cabinets.

zkldi:
i know how the game works and i don't care what konami says or implements. it's absolutely none of my business how networks, people or konami themselves implement internal functions. the types we assert are what is *literally* correct about the charts, regardless of how easy it is to pull from the game.

All charts have a `maxExScore` because `maxExScore` is just a function of the amount of notes in the chart. It does not make sense for this field to be nullable.

zkldi:
This PR seems stale. Please re-open this when it's ready for review.

PR #934, opened 2023-08-16
fix: visual bugs

saintnoodle:
Along side some visual bug fixes, this cleans up a lot of the left over header components that I made redundant and completely restructures the stylesheets hopefully in a structure that's easier to follow. 

I've opted to utilise the filenames _index.scss for stylesheets in folders so those folders can be imported directly as opposed to the messy "components/components" etc... All the Bootstrap related sheets are now in their own folder so its imports don't clutter the main file. The "_style" folder has also been removed as everything useful from those sheets has been included in the new stylesheets.

Authorisation checks have also been added on a handful of components that failed to check this. Delete and edit actions should no longer be visible to users who don't own the scores and sessions they're looking at.


zkldi:
looks excellent.

PR #936, opened 2023-08-17
feat: add matching class to maimai DX

j1nxie:
this PR adds matching classes in maimai DX to Tachi - sort of a ranking system that's available during normal play - players will try to beat other players' scores on randomly chosen charts, with the requirements increasing as their class increases. this system goes from B5-B1, A5-A1, S5-S1, SS5-SS1, SSS5-SSS1 then Legend.

i honestly don't know why i left this out in the first place, but as i start to play more i realize that grinding this kinda helps with consistency and everything and i'd like this to be tracked as well.

the tracking mechanic is probably the same as dans - displaying your peak class.


zkldi:
looks good to me, thanks!

PR #938, opened 2023-08-17
fix: forgot to update page

saintnoodle:
A placeholder prop has been added so the component can smoothly handle falsy values


saintnoodle:
I forgot to fix the gap...

PR #940, opened 2023-08-18
feat: system preference option

saintnoodle:
Here's my proposed solution to prevent the site flashing the wrong theme on load and refresh. Below is the un-minified and readable version of the script that is injected into the head as an inline script on build.

```
const root = document.documentElement;
const	light = "light";
const	dark = "dark";
const	theme = localStorage.theme || (window.matchMedia(`(prefers-color-scheme: ${dark})`).matches ? dark : light);

(() => {
	root.style.setProperty("color-scheme", theme === light ? light : dark);
	root.setAttribute("data-bs-theme", theme);
})();
```

Give it a look and let me know whether this is a suitable solution. I would prefer to load a more robust and extensive script, but Vite really does not want to allow me to load one in as a module separate from the bundle.


client/vite.config.js
```diff
@@ -35,9 +35,10 @@ export default defineConfig({
 				data: {
 					VITE_CDN_URL: process.env.VITE_CDN_URL,
 					TACHI_NAME: process.env.TACHI_NAME ?? "Tachi",
-					VITE_ADDITIONAL_HEAD: process.env.VITE_GOATCOUNTER
+					THEME_INIT: '<script>var r=document.documentElement,l="light",d="dark",t=localStorage.theme||(window.matchMedia(`(prefers-color-scheme: ${d})`).matches?d:l);r.style.setProperty("color-scheme",t===l?l:d),r.setAttribute("data-bs-theme",t);</script>',
```
zkldi, line 38: don't minify the script at all -- you're just obfuscating what the code does

just write out the actual code here. you can use backticks `\`` to have newlines inside your strings.

zkldi (changes requested):
don't minify source code

zkldi:
> ```(() => {
> 	root.style.setProperty("color-scheme", theme === light ? light : dark);
> 	root.setAttribute("data-bs-theme", theme);
> })();```

i don't understand what the point of this IIFE is, whats wrong with just having `root.style...` at the top level of the program?

client/vite.config.js
saintnoodle: My approach was to impact the size of index.html as little as possible. If that's not a concern to you, then we can insert a less compact function into the browser if you want to maintain readability on our end.

zkldi: the difference isn't even a kilobyte, it doesn't matter.

zkldi (approved):

PR #941, opened 2023-08-21
jubeat Ave. support

Meta-link:
Updated seeds to support jubeat Ave. up to 2022122000
Set Ave. as hot game (used to calculate jubility)
Fixed wrong music bar info and canvas size


zkldi:
looks good. thanks

PR #947, opened 2023-09-07
Add Mikado hook as a featured import method for SDVX

adamaq01:
Here's a link of the hook's source code: https://github.com/adamaq01/mikado

On the discord it seemed that some people were interested in it being featured so here's the PR.

There are some places where I am not sure what to put (for example the description of the hook, since theres also Barbatos) so I put some TODOs in there. It would be great to discuss about what these should be.
Same thing for the position on the webpage etc...
Marking it as a draft until this is resolved :)


zkldi:
If this hook supercedes barbatos then maybe you can inherit all of the barbatos descriptions, and barbatos can be marked as deprecated/for older versions of sdvx? I'm not up to date with what works on what anymore. That's what I'd do anyway.

Neat that the source code for this is public. Very interesting to see the rust guts of hooking into DLLs like this :P

PR #955, opened 2023-10-09
new Infinitas pack + monthlies

justOkapi:
Lots of songs to add.  Sorry for the delay, been a rough few weeks.


justOkapi:
Estella is missing from the adds.  I'm investigating, but it looks like there's something fucky with some of the 5th Style ingameIDs?

justOkapi:
Yeah there's a few 5th Style songs that have inGameIDs with format 52xx and I'm pretty sure that's not right but I want to double check that I'm not missing something so I'm waiting on input from anyone else who may have had their fingers in this honey pot.  Once I have a better idea what the hell this is all about I'll get that fixed and add Estella as well.

justOkapi:
JK it is right, these are old songs from prior to the current inGameID schema.  Added Estella's new inGameID and we're off to the races.

zkldi:
old omnimix stuff starts a lot of its CS ports at 5200, and non-cs stuff at 5100. iirc

zkldi:
all looks good to me. thanks!

PR #962, opened 2023-10-15
[iidx] resident final and minor script change

GRIM657:
Updates seed data to 20230905.

The change to convert.ts should make it easier to import new leggendaria charts.  The flag allows partial IFS files to be extracted without their parent, this should be safe because the new .1, which contains all charts, will always be in the new partial file (-p0.ifs files).

### DIFF
--- a/database-seeds/collections/songs-iidx.json
+++ b/database-seeds/collections/songs-iidx.json
@@ -24395,5 +24395,643 @@
 		"id": 2206,
 		"searchTerms": [],
 		"title": "Vector"
+	},
+	{
+		"altTitles": [],
+		"artist": "BEMANI Sound Team \"Dustup Vs. L.E.D.-G\"",
+		"data": {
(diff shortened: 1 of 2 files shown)


zkldi:
Something seems wrong about this PR. Lots of charts here are marked as `isPrimary: false` but aren't superceded by any new primary charts. What's happened?

GRIM657:
> Something seems wrong about this PR. Lots of charts here are marked as `isPrimary: false` but aren't superceded by any new primary charts. What's happened?

Do you have an example of a chart that was deprimaried and does not have a new primary?

The following charts were replaced in resident, and have been deprimaried intentionally.  They should all have new primaries:

MAX 300 SPN
fun DPN DPH
Deep Roar SPA DPN DPH DPA
The Shining Polaris Kors K SPN SPH SPA SPN DPH DPA

PR #963, opened 2023-10-18
fix: extra validator for maimai DX score verification

j1nxie:
checked up with a few friends and 100.5% seems to be the minimum for an ALL PERFECT score, so adding it as a validator for scores.

there isn't a validator for the other way around since scores higher than 100.5% can be not ALL PERFECT.

### DIFF
--- a/server/src/game-implementations/games/maimaidx.ts
+++ b/server/src/game-implementations/games/maimaidx.ts
@@ -146,6 +146,10 @@ export const MAIMAIDX_IMPL: GPTServerImplementation<"maimaidx:Single"> = {
 			if (s.scoreData.lamp !== "ALL PERFECT+" && s.scoreData.percent === 101) {
 				return "A score of 101% should be an ALL PERFECT+";
 			}
+
+			if (s.scoreData.lamp === "ALL PERFECT" && s.scoreData.percent < 100.5) {
+				return "Cannot have an ALL PERFECT without at least 100.5%.";
+			}
 		},
 	],
 };


zkldi:
can you explain the maths behind this? if it's always the case *for now and every future score*, then yeah

beerpiss:
To put simply, the 101% in maimai DX scores is split into 2 parts: 100% base score and 1% break bonus.

All note types (tap, hold, slide, touch, break) award full base scores on PERFECTs, no surprises here.

However, break bonus varies based on how precise a break note was hit, with the minimum bonus for a PERFECT hit being half of full bonus.

Assuming all breaks were hit with minimum bonus, that leaves us with 1 / 2 = 0.5% break bonus => 100.5% score being the minimum for ALL PERFECTs.

This doesn't apply the other way around (100.5+ does not mean an ALL PERFECT) because you could get higher break bonuses and use that to tank the losses from hitting non-perfects.

More details on the maimai DX scoring system: https://github.com/SpiritsUnite/maimai-score-details#maimai-scoring-system

zkldi:
what if a chart has no breaks?

beerpiss:
all official charts are guaranteed to have at least one break, it's been 10 years of maimai and no charts with zero breaks exist.

j1nxie:
to be fair, we support only official charts from SEGA, which will clearly guarantee at least 1 break note, due to the fact that the game checks for 101% for ALL PERFECT+ lamps. in which case, 100.5% being the minimum for ALL PERFECT will also apply.

zkldi:
i think it's more likely that in the 0 breaks case you end up with +1% for breaks no matter what. has anyone actually tested this is what i'm getting at

zkldi:
bad news: no breaks => 100% max. technically speaking maimaidx has a chart dependent max, but do we care? do you care? i don't. sega don't care either. nobody cares.

PR #964, opened 2023-10-25
feat: MAIT Integration

beerpiss:
Add MAIT as an import method for maimai. Since they do not have OAuth, users must manually enter an API token, which is never shown in the client.


zkldi:
is this ready for review?

beerpiss:
> is this ready for review?

Should be now.

beerpiss:
Renamed `MYT` to `MAIT` because `MYT` can be confused for another network.

zkldi:
i don't get this. they provide an API that exports scores in batch manual. Why can't they just post it to `/ir/direct-manual`?

 I don't think I'll accept this PR because the onus is on them to do the rest of the integration here. There's no need for us to lug around code like this when we've done all the work for them.

zkldi:
yeah as far as i'm concerned this is a poor solution to something that isn't our problem. Users shouldn't need to paste this API key in at all - we provide an OAuth2 api that does all of this stuff and they already work in terms of batch manual.

They need to develop this, not us.

PR #971, opened 2023-11-14
feat: support iidx omnimix 1.30.1

GRIM657:
This adds support for resident omnimix, with a few catches:

 - What was originally Scripted Connection Long Mix SPA is now SPL.  This is treated as a new chart in 30-omni.

 - What was originally Junglist King Long SPA and DPA is now SPH and DPH.  This is correct, it was erroneous before, but to maintain compatibility these are also treated as new charts in 30-omni.

 - "Theme from Flo-jack" has bad data and is not supported in 30-omni atm (hoping ryuuou patches this).

 - Deep Clear Eyes has a bug (Anothers and Hypers swapped) and not supported in 30-omni atm (hoping ryuuou patches this).

 - Morning Prayer is I think also bugged (?) and not supported in 30-omni atm.
 
 
 Everything else should (hopefully) be good :)

### DIFF
--- a/common/src/config/game-support/iidx.ts
+++ b/common/src/config/game-support/iidx.ts
@@ -259,6 +259,7 @@ export const IIDX_SP_CONF = {
 		"27-omni": "HEROIC VERSE Omnimix",
 		"28-omni": "BISTROVER Omnimix",
 		"29-omni": "CastHour Omnimix",
+		"30-omni": "Resident Omnimix",
 		"27-2dxtra": "HEROIC VERSE 2dxtra",
 		"28-2dxtra": "BISTROVER 2dxtra",
 		bmus: "BEATMANIA US",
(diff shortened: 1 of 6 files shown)


zkldi:
Yeah this omnimix release is really screwed up. The ScriptedLong changes are particularly painful. I'll have to comb over this and mull over the best way to deal with this.

The solution of splitting charts like this is *technically* correct but really inconvenient for our users. We already do a lot of things to hide implementation details from users (like melding multiple inGameIDs), but as it stands we don't have anything to deal with problems like this.

Rough outline for me to come back to:
- new scriptedlong has 12250; They are customs melded from their normal, hyper, another and black another difficulties, respectively.
- old is 21201; there's only an SP Another and it corresponds to the new SP LEGGENDARIA

Fundamentally tachi doesn't have the infrastructure to handle joining things like this, but it's not *too* different from the ANOTHER->LEGGENDARIA merge we did a *long* time ago. In theory, we could write code that hacks around this and automatically coerces `21201-ANOTHER` into `12250-LEGGENDARIA` (o.e.)

Everything else looks like typical omnimix breakage, all of those charts are troublemakers for (i'm sure) a myriad of reasons.


Blegh. I'll look into this more another time but I *do* want to get 30 omni merged up. Thanks for doing all the work here, I just don't have the time to keep up with this stuff anymore.

GRIM657:
Scripted should be addressed now, like we discussed

PR #973, opened 2023-11-19
Infinitas junk

justOkapi:
Also noticed Bistrover onwards is missing a lot of searchTerms, I started dealing with that then kinda tuckered out.  I'll try to get that sorted in the next few days and throw up a separate PR for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,10 +41,3 @@ failed-tests.log
 
 # bootstrap flag
 I_HAVE_BOOTSTRAPPED_OK
-database-seeds/scripts/rerunners/iidx/okapi-inf.js
-database-seeds/scripts/rerunners/iidx/okapi-inf legg.js
-database-seeds/scripts/rerunners/iidx/iidx-mdb-parse/blacklist.txt
-database-seeds/scripts/rerunners/iidx/iidx-mdb-parse/convert.ts
-database-seeds/scripts/rerunners/iidx/iidx-mdb-parse/merge-mdb.ts	`
-.gitignore
-
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,6 @@ database-seeds/scripts/rerunners/iidx/okapi-inf.js
 database-seeds/scripts/rerunners/iidx/okapi-inf legg.js
 database-seeds/scripts/rerunners/iidx/iidx-mdb-parse/blacklist.txt
 database-seeds/scripts/rerunners/iidx/iidx-mdb-parse/convert.ts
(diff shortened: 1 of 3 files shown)


zkldi:
A couple of these added search terms are redundant - we scan all strings to find the chart:

- Flowers doesn't need "flowers", as it's already in the song title
- Same for opioid peptide.
- Raison D'etre already has `RAISON D'ETRE` in another search term
- Tenjou Tenge seems silly. Does anyone actually call it that? People call it nijuu in my experience
- 1841 華麗なる大犬円舞曲 is called lesbian valse in my experience, not big dog 2


Everything else seems fine.

justOkapi:
Roger that.  I'll go through and revert that stuff.

As far as all the aforementioned 日本語 songs that are just straight up missing search terms, I can work on that tomorrow and roll it into this PR, since unless the search terms are automagically searching for romaji, it's going to be a PITA for anyone who doesn't both have a Japanese IME *and* know how to type with it (which I'm guessing is me and maybe 3 other ばか外人).

GRIM657:
Are the changes to .gitignore meant to be included in the PR?  I think those scripts and the blacklist should stay in the repository, and the ignore file is fine as is.  Looks good otherwise :+1:

zkldi:
Merged by hand; skipped the github UI to solve a merge conflict myself.

Please put your scripts inside the `rerunners` folder and commit them. They are almost certainly of use to other people.

If they are not of use to other people, or only need to be ran once, use the `single-time` folder. Either way I don't think these scripts should be gitignored at all.

justOkapi:
As previously discussed on Discord, those scripts are altered with filepath-specific information for my PC and some other fiddling I wasn't sure was kosher and I didn't want to accidentally kill someone's F: or something.  I've already reverted it in my branch to fix the search terms (aside from that bug with the  - - omni flag) so that rerunner should go up when I merge my branches together.

Is it the smartest way to handle it?  No.  Am I smart?  Also no.  :)

zkldi:
> As previously discussed on Discord, those scripts are altered with filepath-specific information for my PC and some other fiddling I wasn't sure was kosher and I didn't want to accidentally kill someone's F: or something. I've already reverted it in my branch to fix the search terms (aside from that bug with the - - omni flag) so that rerunner should go up when I merge my branches together.
> 
> Is it the smartest way to handle it? No. Am I smart? Also no. :)

This is all fine. If you include your scripts in a future commit I can tidy them up for you and make them generalised. You can then create files like `single-time/foo.js` which call your re-runnable scripts. I can show you how to set all of that up when I'm free.

PR #974, opened 2023-11-20
Add Saekawa hook as an import method for CHUNITHM

beerpiss:
Source code: https://github.com/beerpiss/saekawa

This hook supersedes ChunItachi as an import method.


zkldi:
that failure is a flakey test that i've just disabled for good. Nothing I've done has been able to fix it, so it's going to be skipped for the forseeable future.

PR #976, opened 2023-11-29
feat: more chunithm import methods

j1nxie:
a quick PR to add more import methods for chunithm.

## additions

- aqua/artemis exporter
- myt web interface scraper/exporter


zkldi:
thanks

PR #978, opened 2023-12-01
IIDX metadata update

GRIM657:
- minor omni fix
 - ascii search terms for newer iidx songs
 - iidx tierlist data update

I'd like to get some feedback on the tierlist update, and I still need to review it a bit myself.  The tier data is scraped directly from atwiki thanks to some great scripts by @mlaux.  The plan is to eventually fix it up into a rerunner that can be committed.

### DIFF
--- a/database-seeds/collections/songs-iidx.json
+++ b/database-seeds/collections/songs-iidx.json
@@ -24590,7 +24590,7 @@
 		},
 		"id": 2209,
 		"searchTerms": [
-			"IDENDITY"
+			"IDENTITY"
 		],
 		"title": "[ ]DENTITY"
 	},
--- a/database-seeds/collections/songs-iidx.json
+++ b/database-seeds/collections/songs-iidx.json
@@ -17682,7 +17682,9 @@
 			"genre": "UPLIFTING ETHNIC"
 		},
 		"id": 1607,
-		"searchTerms": [],
+		"searchTerms": [
+			"Coma"
+		],
 		"title": "CoMAAAAAAA"
 	},
 	{
@@ -21398,7 +21400,9 @@
 			"genre": "Hi-歌謡タンゴ"
 		},
 		"id": 1930,
-		"searchTerms": [],
+		"searchTerms": [
+			"TE EXTRANO MUCHO"
+		],
 		"title": "涙のテ・エストラーニョ・ムーチョ"
 	},
 	{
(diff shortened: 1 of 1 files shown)


database-seeds/collections/songs-iidx.json
```diff
@@ -17682,7 +17682,9 @@
 			"genre": "UPLIFTING ETHNIC"
 		},
 		"id": 1607,
-		"searchTerms": [],
+		"searchTerms": [
+			"Coma"
```
mlaux, line 17686: Nice, were the search terms for older songs based on their ticker text or manually entered?

GRIM657: Pulled from ticker text whenever it didn't match the full title, with a couple of manual fixups also (like Coma and Kaijin).

database-seeds/collections/charts-iidx.json
```diff
@@ -9234,14 +9234,14 @@
 			"hcTier": {
 				"individualDifference": false,
 				"text": "11S",
-				"value": 12
```
mlaux, line 9237: It looks like all decimal ratings went down - was this intended? (not really sure how the letter tiers map to the decimal ratings...)

GRIM657: This is part of what I want feedback on.  I have no idea how these values were decided on previously.  I see charts in the current data that are in the same "letter" tier but with different numeric values, I don't know if those were conscious decisions or not.

I'm using this mapping

		case 'S+': return level + .9;
		case 'S': return level + .8;
		case 'A+': return level + .7;
		case 'A': return level + .6;
		case 'B+': return level + .5;
		case 'B': return level + .4;
		case 'C': return level + .3;
		case 'D': return level + .2;
		case 'E': return level + .1;
		case 'F': return level + .0;
		default: return level;

database-seeds/collections/songs-iidx.json
```diff
@@ -24468,7 +24589,9 @@
 			"genre": "FORTUNE OVERTURE"
 		},
 		"id": 2209,
-		"searchTerms": [],
+		"searchTerms": [
+			"IDENDITY"
```
mlaux, line 24593: IDENDITY -> IDENTITY?

I guess it's like this in the game too... I never noticed

GRIM657: No that is absolutely my mistake, that was one of the manual ones.  Thanks!

zkldi:
The code at https://github.com/TNG-dev/Tachi/blob/c8b2ca57d4e75f53cd89a5c4facb647931207e5c/database-seeds/scripts/rerunners/iidx/iidx-pe-kr/parse-iidxkr-data.js#L92-L124

indicates what levels should be mapped to what in the iidx 11 tierlists.

GRIM657:
If the new values look good this should be ready to go

PR #989, opened 2024-02-01
fix(maimaidx): add collab songs + fixes

j1nxie:
## additions
- collab songs from the new zundamon collab (sega decided to release this collab for both jp and intl on the same date)

## fixes
- correct more internal levels (apparently my sheet was missing a few things)
- correct isLatest states for charts before buddies


zkldi:
hottest hotfix in all of the lands

PR #991, opened 2024-02-04
fix: iidx: add scripted connection a mix dph to vanilla 30

GRIM657:
Scripted connection A mix DPH was missing from vanilla resident (but not omni).

I think this is because that song was bugged in launch resident data (leggs replaced hypers) so I probably updated each chart to include 30 by hand and missed the DPH.


zkldi:
Thank you!

PR #1013, opened 2024-03-29
feat: Ongeki support

nyairobi:
Please make a new tag of [rg-stats](https://github.com/TNG-dev/rg-stats) because the latest one is still without Ongeki support.


zkldi:
Heya, this is on the backburner for reviews for the time being. I might get around to this this weekend. Sorry.

zkldi:
MUSTFIX:

- Chart notecount should not be nullable and data must be correct for all charts
- OR: Do away with all notecount information and just store platscore, we'll get around to it later (as we do in our SDVX exscore impl)

- Do away with `lessNaiveScore` and `isHot` and all that stuff.

Other than that this all looks OK!

nyairobi:
> * OR: Do away with all notecount information and just store platscore, we'll get around to it later (as we do in our SDVX exscore impl)

Done that after figuring it would be difficult to obtain "correct" data for all charts.

zkldi:
this looks good, i'll merge this up tonight.

PR #1030, opened 2024-04-09
feat: replace MUI tabs with custom implementation

saintnoodle:
This retains the core functionality whilst deriving the styles from our own style system. As a result, @emotion/styled is also dropped since there is no CSS in JS elsewhere in the project.

Currently, it looks identical apart from the buttons. I've opted for an overlay style as opposed to having the buttons take up physical space inline.

Do flag anything you think may have performance implications, but as far as I'm aware, this is slightly streamlined compared to what's in the library to be replaced. There may be the drawback of potentially missing poly-fills and backwards compatibility, but i don't think there's anything cutting edge here.


zkldi:
rebased onto main

zkldi:
yeah this is awesome, thank you so much!

PR #1039, opened 2024-04-17
implement WACCA import from MYT

cg505:
API support is missing for other games (chuni, maidx, ongeki) for now.

This also requires introducing some build support for grpc, see changes to package.json and src/proto/README.md.


zkldi:
rebased onto main

PR #1046, opened 2024-04-24
Added primary charts for new Infinitas exclusive Leggendarias

MinaAngura:
Apologies to Fluzzarn specifically for including 東京神話.


zkldi:
Why was the notecount wrong for Tokyo Shinwa SPL?

PR #1053, opened 2024-04-28
Fix: sink 2dxtra charts to the bottom for real

GRIM657:
There might be a cleaner way to do this than just duplicating the first if statement... But it works.


GRIM657:
Sorry, I got got by "main" being the default branch for PRs.

PR #1054, opened 2024-04-28
fix: sink 2dxtra charts to the bottom for real

GRIM657:
There might be a cleaner way to do this than just duplicating the first if statement... But it works.


MinaAngura:
Honestly I wouldn't hate if we just put them on their own json, if that's even doable.

zkldi:
it's not doable sadly

zkldi:
wrong HEAD. merge into `main` instead, sorry.

PR #1057, opened 2024-04-30
May Infinitas monthlies (going live on 1 May @ 10:00am JST)

MinaAngura:
Once more, with feeling.


zkldi:
In the future please wait for the update to be out instead of pre-empting it. Just incase anything changes or anything announced was mistaken.

In general let's only add stuff after it's been confirmed to be available in-game.

PR #1058, opened 2024-04-30
[maimaidx] Update to BUDDiES final

beerpiss:
Changes:
- Add songs from the [final update](https://info-maimai.sega.jp/4882/) of BUDDiES
- Update internal levels, mostly for lower difficulties

### DIFF
--- a/common/src/config/game-support/maimai-dx.ts
+++ b/common/src/config/game-support/maimai-dx.ts
@@ -212,7 +212,7 @@ export const MAIMAI_DX_SINGLE_CONF = {
 		universeplus: "UNiVERSE PLUS",
 		festival: "FESTiVAL",
 		festivalplus: "FESTiVAL PLUS",
-		buddies: "BUDDiES"
+		buddies: "BUDDiES",
 	},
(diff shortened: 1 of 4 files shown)


zkldi:
Very curious as to why there were remasters in the db with a level of `0`. Surely that's incorrect?

I know this PR fixes that but there should probably not be things with a level of `0`.

j1nxie:
> Very curious as to why there were remasters in the db with a level of `0`. Surely that's incorrect?
> 
> I know this PR fixes that but there should probably not be things with a level of `0`.

i was responsible for the PR that contained BUDDiES and i'm going to be real with you, i have NO idea why it turned up like that. i didn't get a chance to actually see it break until i actually touched those charts in the game, as i did all the same things as i would in previous updates - run the scripts in correct order: importing the song list, then updating chart internal levels, and so on.

it should not have happened in the first place but i guess it's mainly my bad that i didn't proofread the entire thing before pushing ~~though honestly it's like... 4 charts in 3000+ it's very hard to check lmao~~

PR #1059, opened 2024-05-03
feat: stricter seeds, more validation, etc.

zkldi:
Some surprising errors slipped into the seeds and weren't caught by our existing tests. I've added infrastructure to add custom checks per-game so seeds maintainers can more accurately prevent errors from slipping into the seeds.

CC @GRIM657, @MinaAngura, @beerpiss, @j1nxie, @cg505, and anyone else who is contributing to seeds

Please add some appropriate validation tests to this file, and fix all of the issues highlighted.

# IIDX:

```
[ERR] charts-iidx.json | ON - Get on Beat (55) - DP NORMAL (8cbc49c65e02a4c72eac746ae300cdfc24f84612) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | ON - Get on Beat (55) - SP NORMAL (5fe8688fedae2bb08e0f8a82ce4101500610f248) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | Hirofumi Asamoto (ram jam world) - Junglist King (56) - DP NORMAL (cae0f62390dcbafa71e68197e342fa1bea73802d) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | Hirofumi Asamoto (ram jam world) - Junglist King (56) - SP NORMAL (5e2472c46f86737dd0cb157720a26ed29d632409) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | STONE BROS. - LET THE BEAT HIT EM! BASS MIX (58) - DP NORMAL (7bae78d0ae7ec0c4b04d8eee5c8398ed48f7ed3f) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | STONE BROS. - LET THE BEAT HIT EM! BASS MIX (58) - SP NORMAL (c055985c74a040cf6834ebee693b1c6f56e32203) | Error: versions | Versions array should not be empty as this game uses versions..
[ERR] charts-iidx.json | Sota Fujimori - Look To The Sky(cyber True Color) (590) - DP ANOTHER (eb6c16a73809c81f5fbee10d6b34b052aa257241) | Error: versions | Versions array should not be empty as this game uses versions..
```

# WACCA:

```
[...]

PR #1060, opened 2024-05-04
[iidx] Fixed various issues re: Versions tags

MinaAngura:
-Removed the nonexistent IIDX charts brought up in https://github.com/zkldi/Tachi/pull/1059
-Condensed the myriad entries with multiples of unique Version tags into one entry for that unique Version


MinaAngura:
Well now why's it gone and cleared version tags from things it shouldn't have >:(

PR #1061, opened 2024-05-04
[iidx] Fixed various issues re: Versions tags

MinaAngura:
-Removed the nonexistent IIDX charts brought up in https://github.com/zkldi/Tachi/pull/1059
-Condensed the myriad entries with multiples of unique Version tags into one entry for that unique Version

This time without fucking shit up


zkldi:
Looks good. Thanks.

PR #1063, opened 2024-05-08
[iidx] re-added new Infinitas Leggies

MinaAngura:
Somehow only Concon stuck the landing the first time???


MinaAngura:
+bot diff

PR #1085, opened 2024-05-18
[chunithm] Add versions NEW and NEW PLUS

beerpiss:
For players on MYT.


zkldi:
Shouldn't these folders be marked as inactive? Are people really still playing NEW in 2024?

beerpiss:
> Shouldn't these folders be marked as inactive? Are people really still playing NEW in 2024?

MYT is currently on NEW.

PR #1086, opened 2024-05-18
fix(client): make folder lists more readable

nyairobi:
#1085 is going to make Tunism's folder lists even longer. This is an attempt to make them more readable
Version sort + lexical sort + left-align (center-align makes it chaotic)

Before:
image

After:
image


zkldi:
looks great. thank you.

PR #1089, opened 2024-05-20
feat: always grab oldest score for lamps, instead of unstable

zkldi:


zkldi:
oh also we have to reprocess every pb on the site. lol!

PR #1090, opened 2024-05-22
feat: add inohara as a default client

nyairobi:
https://gitea.tendokyu.moe/akanyan/inohara

### DIFF
--- a/server/src/lib/builtin-clients/builtin-clients.ts
+++ b/server/src/lib/builtin-clients/builtin-clients.ts
@@ -223,6 +223,47 @@ import = '/ir/direct-manual/import'
 # Your Tachi API key
 api_key = '%%TACHI_KEY%%'`,
 	},
+	{
+		clientID: "CXInohara",
+		requestedPermissions: ["submit_score"],
+		name: "Inohara",
+		redirectUri: null,
+		webhookUri: null,
+		apiKeyFilename: "inohara.cfg",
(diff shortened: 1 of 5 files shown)


server/src/lib/builtin-clients/builtin-clients.ts
```diff
@@ -223,6 +223,47 @@ import = '/ir/direct-manual/import'
...
+		apiKeyTemplate: `[Options]
+
+# Whether to enable score submissions
+Enable = true
+
+# Timeout for web requests, in seconds
+Timeout = 3
+
+# Tachi instance base URL
+BaseUrl = ${ServerConfig.OUR_URL}
```
zkldi, line 242: does this really work without quotes around the url?

zkldi (approved):

server/src/lib/builtin-clients/builtin-clients.ts
nyairobi: Why wouldn't it? This is not TOML

zkldi: it's very rare that i see a config format correctly parse unquoted strings

nyairobi: And I think I've never seen an ini dialect that would complain about unquoted values.
You can find the parser n Export.cs:58. It's very basic.

PR #1093, opened 2024-05-28
add links to GAN and NAG in service integration page

cg505:
This change is completely untested!!

I think this was meant to be here since GAN and NAG were split out, but somehow got missed. I just noticed it and it looks like an obvious fix. I did this in the github web editor so I have no clue if it works lol


zkldi:
yup

PR #1100, opened 2024-06-05
feat: add casthour pack 2 songs to Infinitas

GRIM657:
This marks all charts from the 2nd infinitas casthour pack as available in inf, [songlist here](https://p.eagate.573.jp/game/infinitas/2/music/index.html#pac)

It also includes the fixes for Reflection into the EDEN and Strong Jaeger from PR 1045.


zkldi:
yeah whoops, i should've caught that that went to staging and not main. I got rid of staging now to prevent this confusion.

PR #1102, opened 2024-06-06
Update maimai/chuni import descriptions for official site import to not just intl

Xinnoh:
the importers work with jp and cn


zkldi:
Thanks!

PR #1105, opened 2024-06-07
feat: add BPL Pack 2 songs to Infinitas

MinaAngura:
plus the monthlies and CH2 pack that I was beaten to the punch on but I don't think that'll hurt anything


MinaAngura:
+bot diff

zkldi:
Thanks!

PR #1111, opened 2024-06-11
fix: include iidx-mdb-parse file storage in dockerignore

GRIM657:
These folders are temporary file storage from the iidx-mdb-parse script and aren't needed in the docker image.


zkldi:
heh

PR #1118, opened 2024-06-24
fix: react-query, same url, different data

### LINKED ISSUE #1116: sessions/scores misinteraction via cache
imagine if you will the metaphorical react-query equivalent of a cross-map tomahawk in nuketown

saintnoodle:
For games with a single rating algorithm, a session page or the PBs overview page for a given `:game:pt:` would make an identical request, however one would be the data itself, and the other would be the data inside an array. This mismatch causes a crash as the two pages want different data yet are provided with something incorrect when cached from the other page.

We should request an array only if there are more rating algorithms than one, otherwise the identical request is made and cached data is consistent and correctly handled.

The reason I tackled it with this method is because the alternative method of giving the array'd query a unique key will cause unnecessary redundant requests for the same data.

closes #1116


zkldi:
yeah this seems right. thank you for sorting this out, one of those painful issues with trying to keep react-query in check

PR #1126, opened 2024-07-12
feat: rating-based colored text for cells

j1nxie:
as requested by spamblock on discord ([relevant message](https://discord.com/channels/683092550160351261/704764260215029840/1247153275522912277))

this makes rating text color in chunithm, maimai, maimai dx and ongeki colored appropriately to their rating brackets (e.g. a 16.00 score in chunithm will have rainbow text)

edit: this now also has support for sdvx.


client/src/components/tables/cells/MaimaiDXRatingCell.tsx
```diff
@@ -0,0 +1,60 @@
...
+	} else if (rating >= 140) {
+		color = "red";
+	} else if (rating >= 80) {
+		color = "yellow";
+	} else if (rating >= 40) {
+		color = "green";
+	} else if (rating >= 20) {
+		color = "blue";
+	} else if (rating >= 0) {
+		color = "white";
```
zkldi, line 47: why is this one not undefined? (referring to your other comment in MaimaiRatingCell)

zkldi (changes requested):

client/src/components/tables/cells/MaimaiDXRatingCell.tsx
j1nxie: oh, this was an oversight, i did MaimaiRatingCell *after* and i've only thought of it then, i'll patch this once i wake up tmr

PR #1133, opened 2024-07-28
fix: correct some accidental deprimaried IIDX charts

GRIM657:
This just re-primaries some IIDX charts, see the DB Seeds Management link for details.


zkldi:
Did you find out how/why these charts in particular were deprimaried?

GRIM657:
> Did you find out how/why these charts in particular were deprimaried?
> 
> EDIT: This is quite a lot of charts too :(

We would have to ask Fluzzarn, who did the 2dxtra 30 support.  If I had my way, 2dxtra would not be supported on Kamai :)

zkldi:
Ah right, must've been a faulty script or something. Thank you for fixing this!

PR #1136, opened 2024-07-29
fix: do not delete PB after reverting import

### LINKED ISSUE #1129: Reverting an import for a chart hides all your scores of that chart
### Steps to reproduce

1. Have at least one existing score on a chart.
2. Make an import with a new score of that chart.
3. Revert that import.

#### Expected behavior
Only the new score gets reverted.

#### Actual behavior 
It appears as if I haven't ever played the chart, and my NaiveRating drops down (even though it's only supposed to go up). This is my real graph for Ongeki:

Screenshot_20240718_004513

Furthermore, adding another import for that chart fixes it.

### Minimal example

1. Import [test1.json](https://github.com/user-attachments/files/16275531/test1.json).
2. Import [test2.json](https://github.com/user-attachments/files/16275532/test2.json).
3. Revert test2.
4. Go [here](https://staging.tachi.ac/games/chunithm/Single/songs/2100/EXPERT) and note you're not there.
5. Import [test3.json](https://github.com/user-attachments/files/16275533/test3.json).
6. Go back to the chart page and note that the score from test1 has re-appeared.

beer-psi:
closes #1129. fix tested on a local instance using the same reproduction steps as the linked issue.

### DIFF
--- a/server/src/lib/score-mutation/delete-scores.ts
+++ b/server/src/lib/score-mutation/delete-scores.ts
@@ -199,7 +199,7 @@ export async function DeleteMultipleScores(scores: Array<ScoreDocument>, blackli
 	for (const score of scores) {
 		const userHasOtherScores = await db.scores.findOne({
 			userID: score.userID,
-			chart: score.chartID,
+			chartID: score.chartID,
 		});
 
 		if (userHasOtherScores) {


zkldi:
jesus christ it's sort of sad how silly these mistakes are (and how consequential they end up being)

thank you for sorting these two out, they've been massive thorns in my side and I haven't had the time to investigate them myself.

PR #1143, opened 2024-08-08
feat: add myt maimai dx import support

albshin:
Adds MYT maimai DX support

There's currently a bug where UTAGE charts are returning UNSPECIFIED from the backend, not that it matters since we ignore UNSPECIFIED difficulties but please do not merge until this is fixed.

All other errors are due to KT MaiMai DX seeds missing omnimix songs.

image


albshin:
Utage bug is fixed

server/src/lib/score-import/import-types/api/myt-chunithm/converter.test.ts
```diff
@@ -40,7 +40,7 @@ const parsedScore: MytChunithmScore = {
 	},
 };
 
-t.test("#ConvertAPIMytOngeki", (t) => {
```
zkldi, line 43: heh

zkldi:
Looks good to me. Thank you for yr work!

PR #1144, opened 2024-08-08
[maimaidx] Update seeds to 2024/08/08 + Add JP-only songs to BUDDiES+

beer-psi:
The last PR was based on International game data, and so missed quite a few songs that were region locked.


zkldi:
Yep, looks good. Thank you!

PR #1150, opened 2024-08-18
feat: disable score validation by judgement for DDR

Gyoo:
The score validation using judgements is broken due to the absence of the correct hold count if some holds have been missed by the player during the song. This breaks the `stepScore` value and computes a wrong score and invalidates some scores.

Until we have a way to get the actual steps count and holds count for every single chart, this validation function must be disabled.


zkldi:
Thank you. This issue sucks but I can't blame you for staggering into it. I should've noticed this before merging.

PR #1156, opened 2024-08-20
refactor(client): unify user profile components

saintnoodle:
Currently, the UGS info component is used in two different ways. This new component merges the functionality of the two and simplifies the data flow.

Current behaviour fetches the game stats of the requested user whenever the component is created. This generates an unneccessary fetch everytime the component is created and, because the component isn't memoised, creates a second fetch after the component renders.

To fix this, it now chooses how to source data based on conditions:

- If a logged in user is viewing their own profiles, the stats are already available and up-to-date in context, so we can use the context to prevent asking the server.
- For viewing other profiles, it now uses react query.

One semi-important thing to note is that is that both instances currently use different methods to sort game:playtypes. The method I've left in is the oldest, and I'm not sure why the dashboard instance deviated from this; if there's a good reason for that let me know.

### DIFF
--- a/client/src/app/pages/dashboard/DashboardPage.tsx
+++ b/client/src/app/pages/dashboard/DashboardPage.tsx
@@ -1,4 +1,3 @@
-import { APIFetchV1 } from "util/api";
 import { CreateGoalMap } from "util/data";
 import { RFA } from "util/misc";
 import { NumericSOV } from "util/sorts";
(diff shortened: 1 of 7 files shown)


saintnoodle:
The change to the React Query hook is important to verify too. I thought it was strange that an error is thrown inside the callback when skip is true when React Query exposes a method of preventing fetches already. The change favours that option instead of throwing inside the callback.

PR #1162, opened 2024-08-24
fix: FlareSkill PB update

Gyoo:
Adding a new PB merge function to update flare in PB scores. This will also fix the profile ranking.

Some actions need to be done to update existing data : 

- Recreate `enumIndexes` for `scoreData.optional` in scores, since there was a bug in `derivers.ts` that prevented the creation of those
- Update PB merges with the newly updated `optional.enumIndexes`
- Recompute profile ratings

### DIFF
--- a/server/src/lib/score-import/framework/score-importing/derivers.ts
+++ b/server/src/lib/score-import/framework/score-importing/derivers.ts
@@ -77,10 +77,7 @@ export function CreateEnumIndexes<GPT extends GPTString>(gpt: GPT, metrics: any,
 		indexes[key] = index;
 	}
 
-	for (const [key, conf] of [
-		...Object.entries(gptConfig.providedMetrics),
-		...Object.entries(gptConfig.derivedMetrics),
-	]) {
+	for (const [key, conf] of [...Object.entries(gptConfig.optionalMetrics)]) {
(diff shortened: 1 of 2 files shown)


server/src/lib/score-import/framework/score-importing/derivers.ts
```diff
@@ -77,10 +77,7 @@ export function CreateEnumIndexes<GPT extends GPTString>(gpt: GPT, metrics: any,
 		indexes[key] = index;
 	}
 
-	for (const [key, conf] of [
-		...Object.entries(gptConfig.providedMetrics),
-		...Object.entries(gptConfig.derivedMetrics),
-	]) {
+	for (const [key, conf] of [...Object.entries(gptConfig.optionalMetrics)]) {
```
zkrising, line 80: jesus this is a pretty rough bug. will need to merge this and then maybe run some rederivations on the database....

zkldi (approved):

zkrising:
That bug is going to cause some serious trouble and I can't merge this without taking down the site for a minute, as this will change score checksumming, resulting in duplicate scores everywhere. Blegh.

zkldi:
fixed in v3. sorry man for making you wait so long, sincerely.

PR #1163, opened 2024-08-24
feat: support exScore, fast, slow, maxCombo as optional metrics for DDR

Gyoo:
Add support for several optional metrics for DDR :
- Fast/Slow/MaxCombo
- exScore, based on SDVX. Also available as a secondary score and session alg

### DIFF
--- a/common/src/config/game-support/ddr.ts
+++ b/common/src/config/game-support/ddr.ts
@@ -4,6 +4,7 @@ import { ClassValue, zodNonNegativeInt } from "../config-utils";
 import { p } from "prudence";
 import { z } from "zod";
 import type { INTERNAL_GAME_CONFIG, INTERNAL_GAME_PT_CONFIG } from "../../types/internals";
+import {FAST_SLOW_MAXCOMBO} from "./_common";
 
 export const DDR_FLARE_CATEGORIES = z.enum(["CLASSIC", "WHITE", "GOLD", "NONE"]);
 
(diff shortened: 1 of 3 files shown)


common/src/config/game-support/ddr.ts
```diff
@@ -132,13 +146,21 @@ export const DDR_SP_CONF = {
...
+		},
 	},
 
 	sessionRatingAlgs: {
 		flareSkill: {
 			description: "Average of your 10 best Flare Points this session",
 			formatter: FmtScoreNoCommas,
 		},
+		exScore: {
+			description: "Average of your 10 best EXScores this session",
```
zkrising, line 161: I think this is a nonsensical way to rate sessions.

The amount of EX Score you can get on a chart is proportional to notes and not much else. This would just be measuring the average notecount of stuff you've played that session, which isn't really a meaningful signal.

server/src/game-implementations/games/ddr.ts
```diff
@@ -336,9 +358,13 @@ export const DDR_IMPL: GPTServerImplementation<"ddr:DP" | "ddr:SP"> = {
...
 			return DDRFlare.calculate(chart.levelNum, flareLevel);
 		},
+		exScore: (scoreData) => {
+			return scoreData.optional.exScore ?? 0;
+		},
 	},
 	scoreValidators: DDR_SCORE_VALIDATORS,
 	sessionCalcs: {
 		flareSkill: SessionAvgBest10For("flareSkill"),
+		exScore: SessionAvgBest10For("exScore"),
```
zkrising, line 368: Same thing here. I don't think a session metric for EXScore is worth having.

server/src/game-implementations/games/ddr.ts
```diff
@@ -336,9 +358,13 @@ export const DDR_IMPL: GPTServerImplementation<"ddr:DP" | "ddr:SP"> = {
 
 			return DDRFlare.calculate(chart.levelNum, flareLevel);
 		},
+		exScore: (scoreData) => {
```
zkrising, line 361: Likewise, I don't think this is a meaningful metric to calculate scores on, because of ex score being mostly a function of Notes In The Chart.

server/src/game-implementations/games/ddr.ts
```diff
@@ -141,6 +141,25 @@ export const DDR_SCORE_VALIDATORS: Array<ScoreValidator<"ddr:DP" | "ddr:SP">> =
 			default:
 		}
 	},
+	(s) => {
+		const { MARVELOUS, PERFECT, GREAT, OK } = s.scoreData.judgements;
```
zkrising, line 145: I probably should've caught this out when it was first written, but I don't think all of these key names should be capitalized. Maybe I can fix this in the future with a migration.

zkrising (changes requested):

PR #1164, opened 2024-08-25
fix: DDR score validation with judgements using step counts

Gyoo:
- Imported stepCounts and holdCounts from the best source I could find (that doesn't require manually parsing all the charts from game data, I'm not ready for this). Source is mentioned in seeds/scripts/rerunners/ddr/README.md, this is a public spreadsheet.
- Added chart support for `ScoreValidator`
- Reworked score validation using judgements with that newly added chart data


seeds/scripts/rerunners/ddr/README.md
```diff
@@ -28,3 +28,24 @@ File must be stored in the same folder as the script with the name `musicdb.xml`
 
 2) At the last line of the script, eventually change the version if necessary. Current supported versions are `a3` and `konaste` (which are very similar since konaste is based on a3)
 3) Run `ts-node parse-gameData-xml.ts`
+
+# Update stepCounts
+
+1) Download csv files for Single and Double from https://docs.google.com/spreadsheets/d/10NT1VPYV8JHEpi68SBCTApOO5mWEjIAbJP3C5Csthyk/edit?gid=1455718580#gid=1455718580
```
zkrising, line 34: Using this spreadsheet is a non-starter, sorry.

I'm against using community sheets for notecount information because they are *almost always subtly wrong*. We've used them in the past and been bitten again-and-again-and-again. This makes this PR unmergable straight away because I do not trust the accuracy of this data.

I took a look around the sheet and a couple of things raise my eyebrows, like the "Chart Analyzer" in the spreadsheet, which looks suspiciously close the analysis done by simply love. My gut instinct is that this is someone converting charts into `.sm` and then using stepmania as the analysis tool. I can tell you firsthand that the analysers in Simply Love are outright incorrect at times, and are not to be trusted. I can also tell you that  `.sm` is a sticky business with a couple mean edge cases that might not be handled well here.

Either way, I can't merge sheet data like this into the site.

server/src/game-implementations/types.ts
```diff
@@ -164,7 +164,8 @@ export type GPTGoalProgressFormatters<GPT extends GPTString> = {
  * indicating what the error was on failure.
  */
 export type ScoreValidator<GPT extends GPTString> = (
-	score: ScoreDocument<GPT>
```
zkrising, line 167: This is sensible and should be merged regardless of the sheet data issues, although `chart` need not be optional.

zkrising (changes requested):

server/src/game-implementations/types.ts
Gyoo: I'll make a separate PR for that and set this one as draft until I can figure out another solution for stepcounts then

zkrising:
What changed between now and the previous state of affairs? You've done a force-push so I can't exactly figure out what was done myself.

Gyoo:
Sorry, bad habit from work.
Step counts are now parsed from game data instead of that massive spreadsheet I found online

zkrising:
Awesome stuff. Thank you man.

PR #1176, opened 2024-09-17
feat: maidx + chunithm seeds update

j1nxie:
## changes

- chunithm: mark new plus folders as inactive, as mythos got sun (plus?) now.
- maimaidx: seeds update to remove songs with expired license from buddies+.

data from @beer-psi as always.


zkrising:
Another day another dollar. Thanks.

PR #1189, opened 2024-10-03
fix(client): use the rating formatter specified in tachi-common for history graphs

### LINKED ISSUE #1185: Rating graph y axis only displays 00.00 if rating is too high
Seen on DDR profiles : Flare Skill will easily go to 5 numbers and the ratings graph doesn't display correct values on the Y axis.

Example :

beer-psi:
Fixes #1185

Other changes:
- Dropped trailing decimals for MUSECA CuratorSkill. It was always `.00`.
- Changed `TimelineGraph` margins to accommodate bigger y-values 

new rating graph that uses `FormatGPTProfileRating`. the trailing `.00` has been removed.


zkrising:
thank you for fixing this!

PR #1190, opened 2024-10-03
fix: invert folder thermometer data direction

guardianblue:
better lamps/grades should start from bottom,
so that the cumulative lamps/grades can be deduced more easily

Screenshot 2024-10-03 232154
Screenshot 2024-10-03 232208


zkrising:
Yeah this is correct. I'm not sure why this was ever reversed in the first place.

PR #1192, opened 2024-10-08
fix: make-things-available-in-inf: chart.data.inGameID can be an array

GRIM657:
Allow `scripts/rerunners/iidx/make-things-available-in-inf` to check for a match where `chart.data.inGameID` is an array.


zkrising:
Looks good. Thanks.

PR #1193, opened 2024-10-08
feat: enable fer imports for iidx 25, 26, and 26-omni

GRIM657:
This allows fer imports from IIDX 25, 26, and 26 omnimix.


zkrising:
Not passing typecheck? double check imports

GRIM657:
Sorry about that

PR #1198, opened 2024-10-25
Add delta to all folder stat showcases

HutchyBen:
Currently the delta does show for percent, score and playcount on folders however lamp and grade folder showcases does not show.


zkrising:
Good catch. Thanks!

PR #1201, opened 2024-10-30
feat: Epolis Omnimix support

GRIM657:
This adds support for Epolis Omnimix 1.31.1

Carried over from Resident Omnimix, there are two unsupported songs.  I've asked for clarification about Deep Clear Eyes.  Morning Prayer might just need to be counted as a new set of charts.


zkrising:
Another Year Another Dollar. Thanks for your continued support.

PR #1207, opened 2024-11-10
remove newline from MyPageScraper CSV parser check

cg505:
It seems there are CSVs in the wild that use \r\n instead of just \n, which can cause this to fail. Since this check is already best-effort, just drop the newline since it doesn't really matter.

Untested. Unclear if the \r\n can cause additional downstream  serverside parsing issues that will also need to be addressed.

### DIFF
--- a/client/src/app/pages/dashboard/import/WACCAMyPageScraperPage.tsx
+++ b/client/src/app/pages/dashboard/import/WACCAMyPageScraperPage.tsx
@@ -7,7 +7,7 @@ import React from "react";
 function RecordsParseFunction(data: string) {
 	if (
 		!data.startsWith(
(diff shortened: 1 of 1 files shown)


zkrising:
I need to check how csv parsing on the server works first. It should be the case that it handles \r\n and \n the same.

cg505:
We are using default csv-parse options: https://github.com/zkrising/Tachi/blob/38047bbd80f13015394d2481e05b898dc9d53690/server/src/lib/score-import/import-types/file/wacca-mypage-scraper/parser.ts#L55
csv-parse will autodetect the correct record_delimiter: https://csv.js.org/parse/options/record_delimiter/
So it should work but feel free to test.

zkrising:
I trust your judgement with my life and if you are wrong you owe me a pizza when we meet again :pray:

zkrising:
(sincerely yeah this looks good; lets merge)

PR #1208, opened 2024-11-12
Remove duplicate Maimai DX A5 class

Rekunan:


zkldi:
took me 19 months to fix this

PR #1212, opened 2024-12-04
fix: lamp order for ddr

Gyoo:
LIFE4 lamp should not be the highest lamp possible for a song.


RadSwouse:
I just noticed this bug as well, any chance this could get another look?

zkldi:
took me 18 months to fix this

PR #1213, opened 2024-12-04
New match type: `ddrSongHash`

tranq88:
`ddrSongHash` is a 32-character hash used by the official DDR e-amusement website to identify a song.

I believe Tachi supported this match type in the past but I didn't look too much into it. Re-implementing this will allow users to use services that scrape DDR scores from e-amusement, such as 3icecream (however their timestamps can be unreliable; I'm working on my own tool as well).

I tried my best to figure it out without official documentation. If anything's missing, please advise!

### DIFF
--- a/seeds/scripts/test/match-type.test.ts
+++ b/seeds/scripts/test/match-type.test.ts
@@ -31,7 +31,7 @@ const MATCH_TYPE_CHECKS: Record<
 	MatchTypes,
 	| {
 			type: "SONGS";
-			fn: (s: SongDocument) => string | Array<string>;
+			fn: (s: any) => string | Array<string>;
 	  }
 	| {
 			type: "CHARTS";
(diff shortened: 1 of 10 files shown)


seeds/scripts/test/match-type.test.ts
```diff
@@ -61,6 +61,34 @@ const MATCH_TYPE_CHECKS: Record<
...
+			//
+			// unfortunately, they decided to keep the same ddrSongHash across both!
+			// so we manually distinguish it here.
+			const B4UAcolyteSongHash = "01lbO69qQiP691ll6DIiqPbIdd9O806o";
+			if (c.data.ddrSongHash === B4UAcolyteSongHash) {
+				if (c.versions[0] === "a20plus") {
+					return `${c.data.ddrSongHash}-${c.difficulty}-a20plus`;
+				}
+			}
+			// perhaps we could just merge the two "songs" together in the seeds to avoid all this?
```
zkrising, line 87: This has to be handled specially somehow. If we put this in the seeds like this it's a massive landmine for people to step on!

For now, make both of these charts songHash undefined. We'll come back around to fixing it. Not ideal, but not terrible either.

server/src/lib/score-import/import-types/common/batch-manual/converter.ts
```diff
@@ -432,6 +432,46 @@ export async function ResolveMatchTypeToTachiData(
...
+
+			let chart: ChartDocument | null;
+
+			if (context.version) {
+				chart = await db.anyCharts.ddr.findOne({
+					"data.ddrSongHash": data.identifier,
+					playtype,
+					difficulty,
+					versions: context.version,
+				});
```
zkrising, line 446: Shouldn't `songHash` be on a SongDocument instead of a ChartDocument? That makes a lot more sense to me.

zkrising (changes requested):
Needs some small fixes/justifications before I'm willing to merge, but otherwise looks good. Thank you.

nyairobi:
It may be a good idea to mention it in the docs before we forget

* [ddr-SP.md](../blob/main/docs/docs/game-support/games/ddr-SP.md)
* [ddr-DP.md](../blob/main/docs/docs/game-support/games/ddr-DP.md)
* [batch-manual/index.md](../blob/main/docs/docs/codebase/batch-manual/index.md)
* [match-types.md](../blob/main/docs/docs/game-support/common-config/match-types.md)

There are leftover references to the previous `ddrSongHash` implementation, funnily enough

https://github.com/zkrising/Tachi/blob/dc158343d6549977ef42ff22851cc0d96047f4f5/docs/docs/codebase/batch-manual/index.md?plain=1#L75 and https://github.com/zkrising/Tachi/blob/dc158343d6549977ef42ff22851cc0d96047f4f5/docs/docs/codebase/batch-manual/index.md?plain=1#L77

zkrising (approved):

PR #1214, opened 2024-12-10
feat(ongeki): graphs

nyairobi:
This branch adds 3 optional graphs to Ongeki scores:
* Score−
* Missed bells
* Life gauge

The provided data is the Y axis. The X axis is presumed to increase in 1-second intervals. The score provider is trusted with padding the array with `null`s in case of an early forfeit.

Because Ongeki scores are very top-heavy, I capped score− at S (970k), the lowest score that can be considered relevant. Another option could be to scale the Y axis depending on the score.

The scale of the bell chart's Y axis depends on the number of bells that could have been obtained at the time of forfeit. If that's a blocker, one solution would be to add total bell counts to seeds and expose them to the client; alternatively, add another optional field and expect the score provider to fill it every time.

Screenshot_20241210_012833


zkrising:
Thanks for the details. I don't think the bell thing is much of a blocker. Seems like a reasonable limitation to me -- and one that we can fix in the future when we have that information.

This is excellent stuff. Thank you for this!

PR #1154, opened 2024-08-19
feat: songTitle+artist match

nyairobi:
This branch adds an optional `artist` field to the batch-manual spec, which gets taken into account if matching by `songTitle`. Consider it a specialized `identifier2`. I think this is the path of least resistance towards `songTitle` not sucking.

Not benchmarked because I wouldn't know how.

Tested with the import below.
There are 3 Singularities and 1 World Vanquisher in the game.
Everything is imported correctly except the `30` score, as expected, because it doesn't have an `artist` field.

```json
{
    "meta": {
        "game": "ongeki",
        "playtype": "Single",
        "service": "Test"
    },
    "scores": [
        {
            "score": 10,
            "noteLamp": "LOSS",
            "bellLamp": "NONE",
            "matchType": "songTitle",
            "identifier": "Singularity",
            "difficulty": "MASTER",
            "artist": "ETIA.「Arcaea」"
        },
        {
            "score": 20,
            "noteLamp": "LOSS",
            "bellLamp": "NONE",
            "matchType": "songTitle",
            "identifier": "Singularity",
            "difficulty": "MASTER",
            "artist": "technoplanet"
        },
        {
            "score": 30,
            "noteLamp": "LOSS",
            "bellLamp": "NONE",
            "matchType": "songTitle",
            "identifier": "Singularity",
            "difficulty": "MASTER"
        },
        {
            "score": 40,
            "noteLamp": "LOSS",
            "bellLamp": "NONE",
            "matchType": "songTitle",
            "identifier": "World Vanquisher",
            "difficulty": "MASTER"
        },
        {
            "score": 50,
            "noteLamp": "LOSS",
            "bellLamp": "NONE",
            "matchType": "songTitle",
            "identifier": "World Vanquisher",
            "artist": "void (Mournfinale)",
            "difficulty": "MASTER"
        }
    ]
}
```

PR #1104, opened 2024-06-06
feat: make the duplicate-song-title-checker work

zkldi:
I made the duplicate-song-title-checker work, but it's revealed that we have a hell of a lot of duplicate song titles in the database. This needs to be sorted out one way or another.

We have a fairly frustrating issue which is that the eamusement CSV imports use song title matching, but they *can't* because there are duplicate song titles in the games-in-question. This sucks, we need to do this abstraction better. Sad.

PR #1218, opened 2024-12-11
docs: document #1154 (songTitle+artist matching)

nyairobi:
> Dany kind of forgot about documentation.

\>\>#1154

* Documented the `artist` identifier.
* Made the `songTitle` warning less scary because duplicates have been being resolved properly since #1104.
  * Almost all KT GPTs enable this method so a super scary warning isn't a good look in the first place.
 * Updated the Ongeki doc.


zkrising:
Thanks. Looks good.

PR #1219, opened 2024-12-11
feat: additional classes for gekichumai

nyairobi:
This branch adds additional classes for Ongeki (16.00, 17.00), Chunithm (17.00) and maiDX (16k). Marking this as a draft because Chunithm and maiDX maintainers have to be on board, and there are details to discuss. The names of the new classes are basically placeholders. This is just a proposal that has been preliminarily discussed in the #general chat.

pic1

The motivation for these classes is "they are cool", and can be something to strive for. In all three games, achieving rainbow is generally considered easy as long as you play seriously, and definitely can't be considered endgame. Ongeki in particular has experienced severe rating inflation, hence there is also "Rainbow +" at the halfway point. I'm still not sure about that one.

The term 虹(極) and the purple/blue gradient is an official designation for achieving combined 50k rating in all three games (for example ,17.00 Ongeki + 17.00 Chunithm + 16k maiDX). As for the English name:
* "rainbow extreme" is a literal translation of 虹(極). 
* "rating master" is a nameplate received for achieving rating 17.00 in Ongeki.
* "true rainbow" is a term some people use. 

Things to consider:
* Whether this is a good idea in the first place.
* Whether to call it 虹(極) or just 極 for consistency, and then there's maimai.
* What to call it in English (see above).
* Whether it would be a better idea to put it a new separate enum, just in case S*GA decides to shake things up (unlikely but possible) .
  * If not, I don't know how it would affect the existing data.
* Whether to include maimai classic.

Finally, to back this up with data, our current leaderboards are as follows:

  | Chunithm | Ongeki | maiDX
-- | -- | -- | --
≥ 17.00 | 20 | 1 | *N/A*
≥ 16.00 | 147 | 25 | 44
≥ 15.00 | *N/A* | 60 | 156
≥ 10.00 | 478 | 131 | 313

10.00 has been chosen semi-arbitrarily for counting the number of players (players below 10.00 often don't have enough scores to fill the naiverating window). Percentages of players against that 10.00 baseline is as follows: 
  | Chunithm | Ongeki | maiDX
-- | -- | -- | --
≥ 17.00 | 4.18% | 0.76% | *N/A* <sub><sup><sub><sup>literally impossible</sup></sub></sup></sub>
≥ 16.00 | 30.75% | 19.08% | 14.06%
≥ 15.00 | *N/A* | 45.80% | 49.84%

pic2

PS Turns out Chunithm actually introduced this purple gradient for rating 17.00 *as I was writing this comment*

### DIFF
--- a/client/src/lib/game-implementations.tsx
+++ b/client/src/lib/game-implementations.tsx
@@ -62,6 +62,11 @@ export const GPT_CLIENT_IMPLEMENTATIONS: GPTClientImplementations = {
 						"linear-gradient(-45deg, #f0788a, #f48fb1, #9174c2, #79bcf2, #70a173, #f7ff99, #faca7d, #ff9d80, #f0788a)",
 					color: "var(--bs-dark)",
 				},
+				RAINBOW_EX: {
+					background:
+						"linear-gradient(-45deg, #0fa091, #0f98d5, #67087f, #d9007e, #f56e06)",
+					color: "var(--bs-light)",
+				},
 			},
 			dan: {
 				DAN_I: bgc("#3b66f1", "var(--bs-light)"),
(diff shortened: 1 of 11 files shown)


nyairobi:
Since Chunithm Verse has somewhat unexpectedly rescaled difficulties and introduced a new color for 17.00, this change is now bound to happen for Chunithm, and has to be put on hold for the Ongeki and maiDX

PR #1233, opened 2025-01-11
feat: support iidx 31-2dxtra

GRIM657:
This adds support for Epolis 2dxtra, and also shoehorns in some altTitles for Pinky Crush CSVs.

This is my first time adding a 2dxtra version, but it looks like everything went well?  Thanks for making it easy :)


GRIM657:
Ah, nevermind, I should have re-started the server locally and caught these issues.  I think the deprimary logic isn't working for cases where a song's 2dxtra inGameID has changed due to infinitas ports etc.  I'm going to mark this WIP for now.

GRIM657:
OK we should be good now :)

Also, there were three unreleased charts in 32 that were revealed like the day after I sent in the PR lol.  So I slipped those in too.

PR #1243, opened 2025-02-17
fix(ongeki): restore previous folders

nyairobi:
The folders should have been copied rather than replaced.


zkrising:
Yeah. I should've caught this. Sorry.

PR #1257, opened 2025-03-20
fix(client): streamline song-chart info

nyairobi:
* Made Ongeki use the IIDX song info format because I think it looks better
* Deduped BMS/PMS/IIDX/pop'n logic
* Removed ITG

Any game with `song.data.genre` could potentially use the IIDX format

Also, naming was inconsistent between `SongInfoFormat` and `SongChartInfoFormat`


client/src/components/game/songs/SongChartInfoFormat.tsx
```diff
@@ -0,0 +1,55 @@
...
+								levelNum={chart.levelNum}
+								prefix="Internal Level: "
+							/>
+						</h6>
+						<h6>
+							<Muted>
+								From オンゲキ{" "}
+								{(chart as ChartDocument<"ongeki:Single">).data.displayVersion
+									.replace("オンゲキ", "")
+									.trim()}
```
zkrising, line 45: Why not just from {.. displayVersion}. What's the replacement/trimming for?

nyairobi: displayVersions are like `'オンゲキ', 'オンゲキ PLUS', 'SUMMER', 'SUMMER PLUS', 'R.E.D'`, and so on.

The replacement turns that into `'オンゲキ', 'オンゲキ PLUS', 'オンゲキ SUMMER', 'オンゲキ SUMMER PLUS', 'オンゲキ R.E.D'` etc.

This, by all means, could be adjusted in the seeds instead, as I believe displayVersion isn't used anywhere else

zkrising: Ah, it's whatever then. This is fine enough.

PR #1260, opened 2025-03-26
seeds: Ongeki Act.3 final

nyairobi:
Also added chart view links, and a quick and dirty script that fetches them.


zkrising:
Frustrating that the chart view URLs can't be automatically determined from existing data. I'm baffled as to why these chart viewer websites have stolen the worst part of textage (having a random, underivable html filename).

Either way this all looks good. Thanks for your work.

PR #1261, opened 2025-03-27
CG Jubeat Import

defvs:
CG API now returns `musicRate`, all barriers to implementation are lifted unless there are some nasty bugs in the CG API.

image

Draft because the API call still needs to be checked on CG's side and all enums-to-int mappings confirmed; I've DM'd Radio about this.

Checklist:
- [X] Server implementation
- [ ] Validate/correct API Ints to Enum mappings
- [X] Client implementation
- [X] Backend testing
- [ ] Frontend testing
- [ ] Discord bot implementation


defvs:
- Frontend now seems to work fine
- Enums-to-int are correct
image

zkrising:
Looks excellent. Thank you for this work and for being patient.

PR #1263, opened 2025-03-30
feat: O.N.G.E.K.I. Re:Fresh rating algorithm

nyairobi:
### ScoreRating and StarRating

These are the two independent new score rating algorithms that have already been merged into `rg-stats`. 
* ScoreRating is derived from score and lamps, and caps at 18.400 for the hardest chart.
* StarRating is derived from p-score, and is usually worth around 1.000 or below.

### NaiveRatingRefresh

The new in-game profile rating algorithm is
```
(Average of 10 ScoreRatings from the current version, each divided by 5) + (Average of 50 ScoreRatings from previous versions) + (Average of 50 StarRatings)
``` 

The naive version is
```
(Average of 60 ScoreRatings) * 1.2 + (Average of 50 StarRatings)
```
to maintain the ratio between ScoreRatings and StarRatings.

Notably, this rating caps at over 22.000, on a different scale than the score algorithms, hence the distinct name.

### Class value updates

[Reference.](https://files.catbox.moe/pj2fqx.png) Notably, introducing RAINBOW_EX (like chunithm) and RAINBOW_SHINY (new).

### All Break+

This lamp has been introduced in Re:Fresh, and is awarded for scoring the theoretical value (1,010,000), therefore it also implies FULL BELL. It has to be applied retroactively, which is a breaking change that warrants a migration.

### Platinum Score changes

Re:Fresh enabled Platinum Score for all difficulties (previously it was exclusive to the two highest difficulties). Ongeki-Net now shows Platinum Score data. Platinum Score now determines the number of stars which is used to calculate StarRating. The number of stars is what actually matters, and it's something to set goals for, so it should be a derived metric.

Because new metrics can't be derived from optional metrics, and because Platinum Score is no longer situational, it should be promoted to a provided metric, but that's a breaking change.


nyairobi:
Marking this as ready because I've added everything I wanted to add, with two caveats:
* I am clueless in regards to the MYT integration.
* So far I could only test it using my own scores, because the KT anon dataset is too old.

nyairobi:
Rebased, removed `rbreak`, and renamed RatingV2 to ScoreRating, probably making it more consistent:
* ScoreRating, because it's based on score
* StarRating, because it's based on platinum stars
* NaiveRatingRefresh, because it's an amalgamation of the above

The session algorithm is called "naive" for ScoreRating but not for StarRating, because StarRating normally doesn't distinguish between "old" and "new" songs.

nyairobi:
Added an insight page and platinum score graphs, and re-tested the migration.

zkrising:
I'll have some time this weekend. If you can sort out the merge conflicts I'll try and get it merged for sunday. Thank you for your patience.

nyairobi:
Have you reviewed the branch though?

zkrising:
to the best of my ability? I don't play this game; everything here looks correct and aligns with what your summary says in the PR.

PR #1265, opened 2025-04-01
fix(client): "e is undefined" in session search

nyairobi:
This simple nullish check prevents `"error": "e is undefined"` when typing anything in a session search bar. Admittedly this is a band-aid fix, as the PB data is still missing, but at least it doesn't crash.


zkrising:
Thanks, can't believe this has gone unnoticed for so long.

PR #1269, opened 2025-04-04
[chunithm] split lamp into combo lamp and clear lamp

beer-psi:
CHUNITHM has always had a concept of separate clear lamps and combo lamps, since the nonsensical clear and skill system means that you can get an ALL JUSTICE while technically failing. However, since before I took over maintaining the game, the two kinds of lamps have been merged into a single lamp column.

While this wasn't a problem in the past, since CHUNITHM SUN+, harder clear lamps have been introduced, ranging from HARD (20 miss clear) to CATASTROPHY (10 justice or under clear). While we initially ignored it, there's community interest for tracking those lamps, and for better score "correctness", this is an attempt to migrate to the proper lamp system.

This PR contains a database migration to move scores over, as well as updated code that handles those lamps.

The database migration affects anything that might touch lamps: scores, personal bests, stat showcases, and goals. Please comment if you think anything is missing. It has been tested on the Kamaitachi 2024-05 dataset, though CHUNITHM on Kamaitachi has gotten significantly larger since then.

score with CATASTROPHY clear
score with FULL COMBO HARD clear
score with ALL JUSTICE CATASTROPHY clear


zkrising:
Interesting stuff. I'm definitely willing to merge this, just let me know when its ready.

beer-psi:
After a few dataset loads and test migration runs, this should be ready for review. Again, I'm not entirely sure if I've covered all the places where lamps could show up. Currently the migration affects scores, PBs, stat showcases and goals.

beer-psi:
Oops, I think I screwed up folder goals. Back to the mines.

beer-psi:
we're so back

image

beer-psi:
Renamed `comboLamp` to `noteLamp` for consistency with O.N.G.E.K.I.

server/src/lib/migration/migrations/chunithm-lamp-split.ts
```diff
@@ -146,6 +146,7 @@ function MoveLamps(
 	} else if (
 		oldLamp === "FULL COMBO" ||
 		oldLamp === "ALL JUSTICE" ||
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
zkrising, line 149: ESLint's heuristic is poor here, but what it's saying is correct. This condition is unecessary because all other possible states have been exhausted.

With a more simple example, what you are saying is the same thing as `if (cond !== true && cond === false)`; After the previous conditions there is only one possible value it can be, so the check is redundant.

However, I am completely with you here that this check should be maintained and ESLint is wrong for flagging this as an issue. It's just a very poor quality linter.

zkrising:
going to merge this later today hopefully it doesn't blow up the site

PR #1274, opened 2025-04-08
fix(server/score-import): jubeat@CG bux fixing

defvs:
Fixes a nasty edge case: Playing with EXC modifier (fails if you don't do an EXC) means you can get a score saved with no misses (classifies as a FC) but with < 700k score (should be a FAILED).

WIP


defvs:
Superseded by #1276

PR #1276, opened 2025-04-13
fix(api-cg/jb): use clearFlag bitfield to generate Lamp

defvs:
Those bitfield values are taken from what the game sends - Radio told me that this is supposedly the same as what the CG API returns. **I cannot confirm it. I do not have CG API access. Please test this outside of prod.**.

I got the bitfield values from actually dumping the network calls of my own cab, there's further stuff in the MSB of the bitfield but these seem mostly useless.

Here are the bit positions:

| 31:4             | 3   | 2  | 1     | 0            |
|------------------|-----|----|-------|--------------|
| ? | EXC | FC | CLEAR | FAILED/SAVED |

I've created various tests for that as well - including the edge case of a 'challenge fail'.


zkrising:
fuck it, lets see if this works in prod. It's on me if it doesn't.

PR #1277, opened 2025-04-16
feat: update konaste seeds

whiteou7:
Added 7 new songs (konaste exclusive as of now but some of them are available on arcade. I didn't add them as eg charts because of current datacode) - 4 pekora songs, 2 IIDX pack songs and 1 new song
Added konaste version of 8 songs - 8 IIDX pack songs
Added konaste version of XCD charts of 20 songs - XCD effect pack

Note: This is my first time contributing. Everything passed seeds test still.


whiteou7:
did it fail due to bms course lookup

zkrising:
Yeah. That's fixed upstream so this is still good to merge.

PR #1278, opened 2025-04-19
[MYT] Do not check access code prefixes

beer-psi:
This PR removes the client-side requirement that MYT access codes must start with a 0 or a 3, because all FeliCa NFC devices (including stuff like headphones) can be scanned on the network, and the access codes generated for them might not start with a 0 or a 3.


cg505:
thanks

PR #1284, opened 2025-04-25
feat(client): chunithm graphs, song durations

nyairobi:
screenshot

Re-purposed the Ongeki score and life graphs for Chunithm. Aside from timestamps, the picture contains real data. I went with S+ baseline because Chunithm scores are extremely top-heavy.

### DIFF
--- a/client/src/components/charts/GekichuScoreChart.tsx
+++ b/client/src/components/charts/GekichuScoreChart.tsx
@@ -11,7 +11,7 @@ import {
 	PointTooltipProps,
 	LineSvgProps,
 } from "@nivo/line";
-import { COLOUR_SET, Difficulties } from "tachi-common";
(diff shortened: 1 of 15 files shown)


client/src/components/charts/GekichuScoreChart.tsx
```diff
@@ -22,21 +22,39 @@ const formatTime = (s: DatumValue) =>
...
-	switch (s) {
-		case 970000:
-			return "S";
-		case 990000:
-			return "SS";
-		case 1000000:
-			return "SSS";
-		case 1007500:
-			return "SSS+";
+const scoreToLamp = (game: Game) => (s: number) => {
```
zkrising, line 25: This is a scarily misleading function name given what the function does. `getScoreYAxisNotch` or something like that would be better.

PR #1288, opened 2025-04-30
SDVX: add MAXXIVE CLEAR lamp

pinapelz:
Adds MAXXIVE CLEAR lamp that is new to SDVX AC version. 
- Added mapping for this lamp in eamuse sdvx CSV imports

I seperated out USC_LAMPS in the tests since I don't think MAXXIVE is a thing in usc yet and I don't want to mess up IR code by trying to add it there.

Testing with locally built rg-stats:
image

Not a maintainer so please change/fix if I missed something


zkrising:
note to self: this requires a database migration before merging

zkrising:
Please update this branch to be from the current main, as it contains the rg-stats upgrade you need.

zkrising:
Alright: haven't forgotten about this, just been extremely busy. I need to write a migration for this before I can merge it, as it changes lamp values in the database; that's why it's held up. Sorry.

pinapelz:
No rush, don't think many people will have access to this update regardless. US cabs were randomly rolled back to before this update for no apparent reason (with no acknowledgement from KONAMI either, been like this for about a month now)

zkrising:
> It seems the lamp index in the game is actually after PUC, I guess if we do the same here it would avoid having to do a migration. Maybe you could adopt the same strategy here? Though it's a bit weird to have it defined in that order 🤷‍♂️

Wouldn't work sadly; the indexes are used for ordering in a lot of places, so this would imply MAXXIVE > PUC.

blobdash:
Would be a good idea to also add the lamp to the converters for imports (e.g. in `server/src/lib/score-import/import-types/common/api-cg/sdvx /converter.ts`, where it should be added to `ConvertCGSDVXLamp` (maxxive is 6)) so when this is merged imports actually work right away.

lmk if you want me to PR it to your branch.

PR #1290, opened 2025-05-01
fix: graph errors

nyairobi:
The lack of song data goes deeper than I thought. I just sort of gave up and added a query.

Additional slight improvement: the header no longer shows up if no charts are available.

### DIFF
--- a/client/src/components/tables/dropdowns/components/OngekiScoreDropdownParts.tsx
+++ b/client/src/components/tables/dropdowns/components/OngekiScoreDropdownParts.tsx
@@ -30,14 +30,14 @@ export function OngekiGraphsComponent({
 		score.scoreData.optional.totalBellCount !== undefined;
 
 	if (!available) {
-		return Box("No charts available");
+		return <Box message="No charts available" />;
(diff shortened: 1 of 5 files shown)


client/src/components/tables/dropdowns/components/OngekiScoreDropdownParts.tsx
```diff
@@ -101,3 +103,16 @@ function GraphComponent({
...
+		<div className="col-12">
+			<div
+				className="d-flex align-items-center justify-content-center"
+				style={{ height: "200px" }}
+			>
+				<span className="text-body-secondary">{message}</span>
+			</div>
+		</div>
+	);
+}
```
zkrising, line 118: This should be

```
function Box({ message }: { message: string }) { Same Function Body }
```
and called as
```
<Box message="your message here" />
```

nyairobi: Sorry, you can clearly tell I'm not a Reactor.

PR #1306, opened 2025-06-18
feat(maimaidx): graphs and song durations

beer-psi:
Graph support for maimai DX's percent and life was added based on nairobi's work in #1214 #1284.

Only one graph data (either percent or life) needs to be submitted for graph to show up. This is because maimai DX life gauge is either a challenge or a track skip option that needs to be enabled explicitly.

~~The PR is feature-complete; currently in draft because I need to find duration data for removed songs.~~

percent graph
life graph


zkrising:
I simply have no time to deal with complex PRs but I promise I will get around to it and appreciate your effort.

PR #1328, opened 2025-07-15
dev: make some bootstrapping steps always run

beer-psi:
Sometimes, due to circumstances, you may have to rebuild the dev container, which means that `node_modules` and installed packages will be lost, but since `BOOTSTRAP_OK` still exists, they won't be reinstalled automatically.

This PR changes the bootstrapping script so that some steps become idempotent and thus we don't need to check for the `BOOTSTRAP_OK` file:
- `mvExampleFiles` uses `cp --update=none` to make sure that the example file won't be copied over if the config file already exists.
- `selfSignHTTPS` checks if the cert files already exists and has not expired before continuing.
- `pnpmInstall`/`installLocalDebs` doesn't do much if the packages have already installed.

### DIFF
--- a/dev/bootstrap.sh
+++ b/dev/bootstrap.sh
@@ -37,8 +37,8 @@ function mvExampleFiles {
 }
 
 function selfSignHTTPS {
-	if [ -e server/cert/key.pem ] && [ -e server/cert/cert.pem ]; then
-		echo "HTTPS Certificates for local-dev server already exists."
+	if [ -e server/cert/key.pem ] && [ -e server/cert/cert.pem ] && openssl x509 -checkend 0 -noout -in server/cert/cert.pem; then
+		echo "HTTPS Certificates for local-dev server already exists and has not expired."
 		return 0
 	fi
 


dev/bootstrap.sh
```diff
@@ -22,21 +22,26 @@ function setupShell {
...
-	cp bot/example/example.env bot/.env
+	cp --update=none bot/example/conf.json5 bot/conf.json5
+	cp --update=none bot/example/example.env bot/.env
 
 	echo "Moved!"
 }
 
 function selfSignHTTPS {
+	if [ -e server/cert/key.pem ] && [ -e server/cert/cert.pem ] && openssl x509 -checkend 0 -noout -in server/cert/cert.pem; then
+		echo "HTTPS Certificates for local-dev server already exists and has not expired."
```
zkrising, line 41: No need to even echo anything here, it doesn't matter.

dev/bootstrap.sh
```diff
@@ -118,6 +123,6 @@ The existence of this file stops Tachi from bootstrapping again.
 There's nothing harmful about this -- you can bootstrap as much as you want!
 We just don't want to necessarily bootstrap *each* time we boot Tachi.
 
-To bootstrap again (incase you think something has gone wrong)
-Delete this file and re-run npm start.
+To bootstrap again (in case you think something has gone wrong)
```
zkrising, line 126: i'll die on my hill of repeatedly mispelling things and never learning

PR #1332, opened 2025-07-20
fix: update sdvx titles/artists

ypsvlq:
sync with game

TOYBOX CANNON doesn't display clearly in GitHub diff, but eamuse site and other sources have an extra combining character, which looks wrong when viewed on Tachi.


zkrising:
What the fuck are these song titles

PR #1341, opened 2025-08-03
feat: merge sdvx2 and sdvx3 mdb

ypsvlq:
booth players rejoice


ypsvlq:
サイバーサンダーサイダー grv would probably have been rerated in sdvx4, i think the rest are okay though

PR #1348, opened 2025-08-17
feat: add minimumRelevantValue and minimumScores to classes

nyairobi:
I removed the if-else chains from `bot/src/webhookHandlers/classUpdate.ts`; the discord bot now depends on optional fields from `tachi-common` to determine whether a class value is worth rendering.

### minimumRelevantValue

> The ID of the lowest class value that should trigger a discord message.
> Lower values will be ignored to reduce channel spam.

This has been added to some existing GPT configs based on the previous contents of `ShouldRenderUpdate()`:
* sdvx: `IMPERIAL_I`
* popn: `GOD`

And a few new ones:
* ongeki: `RAINBOW`
* wacca: `RAINBOW`
* ddr: `URANUS`
* usc: `IMPERIAL_I` for parity with sdvx
* maimaidx: `BRONZE`

### minimumScores

> What number of scores is needed to "fill" this class?
> If the number of scores is lower, new class values will not
> trigger discord messages (useful if minimumRelevantValue is undesired)
> For example, a new SDVX player playing their first 50 charts will blitz through
> at least 10 of the volforce ranks, which would just result in channel spam.

This has been added to maimai, chunithm, gitadora, jubeat based on the previous contents of `GetMinimumScores()` (bms, pms and iidx are dan-only). 
I think `minimumRelevantValue` is better in all cases, but I can't dictate which classes are "irrelevant" in games I don't maintain.

### DIFF
--- a/common/src/config/game-support/maimai-dx.ts
+++ b/common/src/config/game-support/maimai-dx.ts
@@ -204,7 +204,7 @@ export const MAIMAI_DX_SINGLE_CONF = {
 		colour: {
 			type: "DERIVED",
 			values: MaimaiDXColours,
-			minimumScores: 50,
+			minimumRelevantValue: "BRONZE",
 		},
 		dan: {
 			type: "PROVIDED",


zkrising:
Am i merging this? Or are you doing the thing you mentioned in #dev just now?

nyairobi:
It is done

PR #1352, opened 2025-09-04
seeds: Update DDR up to 2025061002

auxbh:
- 68 new songs
- Fixed `flareCategory` that wasn't set properly for some existing songs


zkrising:
Slipped by me, sorry.

PR #1354, opened 2025-09-07
Make flare calculation go through all scores as PB may not have highest flare

HutchyBen:


HutchyBen:
I initially thought there could be some sort of PB merge for best flare but I wasn't too sure how to go about that, I'll have another look now and if I cant figure it out i'll just close this PR and leave it as an issue for another time.

beer-psi:
SDVX has a merge for best exScore, so you can probably just do that?

https://github.com/zkrising/Tachi/blob/main/server/src/game-implementations/games/sdvx.ts#L43-L45

HutchyBen:
Yeah, I think when I was initailly looking at doing a PB merge I was trying to merge flareSkill and not the actual flare which is a bit silly thinking back now. I'll test it out merging the flare in a little bit

HutchyBen:
this has been a dupe of #1162 this whole time teehee

PR #1357, opened 2025-09-16
fix(devcontainer): switch to debian 13

### LINKED ISSUE #1351: Docker file not building correctly with VS Code
After making a fork of the repo, on Step 3, VSC throws "EPERM: operation not permitted, chmod '/tachi/node_modules/tap/bin/run.js'" error at me.  

a bit of "sudo" shenanigans later and i got it to run, but it's not great when it crashes and the only info on the "fix" is that there's no info.

beer-psi:
Update the devcontainer setup.
- Ubuntu 24.10 -> Debian 13
- Silence pnpm update banners
- Potentially fix #1351

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,29 +1,35 @@
 # For use via `devcontainer.json`. The docker-compose-dev file sets some other
 # important variables.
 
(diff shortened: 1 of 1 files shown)


zkrising:
they call him beer "4 prs" psi

PR #1365, opened 2025-09-27
fix: seeds corrections for iidx 32 and 33

GRIM657:
This addresses a few issues:

- Mid-game version 32 removals that were not marked as removed in seeds (they were removed via server flags, data is unchanged, so it's a manual process)
- Missing songs from versions 30-32 in version 33 (issue on my end)
- Missing leggendaria charts (iidx-mdb-parse was sometimes preferring original files instead of patch files, if a -p0 exists it is not just for spl/dpl charts, it is a replacement of the original .1 file)
- I forgot to add the new version of ADVANCE via the "add new primary" script.

I adjusted the logic in iidx-mdb-parse to get the desired result but I think that whole section was made with the assumption that "p0" files are *only* for spl/dpl charts rather than patches to the whole .1, so it could use some further rework.


GRIM657:
Pushed a fix for the seeds error, incorrectly inputted the version in the script for ADVANCE.  Didn't catch it since I didn't restart the server locally, mb

PR #1369, opened 2025-10-05
fix(client): consider pagination in import analyzer entry fragment key

chown2:
Fixes elements with same index in the page staying open when listing pages.

1) Open https://boku.tachi.ac/utils/imports
2) Expand first score
3) List to the second page
4) Observe first score of the page being open

### DIFF
--- a/client/src/components/tables/components/TachiTable.tsx
+++ b/client/src/components/tables/components/TachiTable.tsx
@@ -170,7 +170,9 @@ export default function TachiTable<D>({
 					<tbody>
 						<NoDataWrapper>
 							{window.map((e, i) => (
-								<React.Fragment key={i}>{e && rowFunction(e)}</React.Fragment>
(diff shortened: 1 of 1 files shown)


client/src/components/tables/components/TachiTable.tsx
```diff
@@ -170,7 +170,9 @@ export default function TachiTable<D>({
 					<tbody>
 						<NoDataWrapper>
 							{window.map((e, i) => (
-								<React.Fragment key={i}>{e && rowFunction(e)}</React.Fragment>
+								<React.Fragment key={i + 10 * (page - 1)}>
```
beer-psi, line 173: This looks like it'd break for page sizes different than 10.

beer-psi, line 173: Hardcoding to `10` would break for different page sizes. You should probably use `pageLen` instead.

beer-psi (changes requested):

PR #1430, opened 2026-02-12
[sound voltex] add divine dans / divine skill levels

GlassGrimoire:
add support for divine dan courses from sound voltex
colours are the same as the non divine equivalent
{D372D88B-35CB-45FC-9E32-8EF5F8253627}
{7DBE22A1-B914-417E-8DCB-69F0992F6C19}


beer-psi:
This needs a database migration since the indexes for the non-divine dan levels have been shifted. Probably will just be delayed to Tachi3.

PR #1461, opened 2026-04-20
[chunithm] Update to X-VERSE-X

beer-psi:
- Added versions X-VERSE-X and International/Omnimix variants
- Regular seed changes (songs, charts, folders, tables)


DarinMao:
beer

PR #1478, opened 2026-05-18
feat: update arcaea for v3 + 6.14 seeds/data

nyairobi:
Arcaea support works great. Most changes are related to v3, and some are generic.


nyairobi:
This is now in a really good shape. There is the Eternal diff, 12th dan, latest seeds and per-chart display version with the IIDX-style song info cell. I'm done adding things, and I leave import methods to someone else.

The only thing I dislike is that Beyonds with audio override (Singularity VVVIP and other AF songs, Axium Divergence, Pregnant Resurrection etc.) need a separate song entry but there's probably no way around that.
Maybe the actual April Fools charts (FTR) could be added with level 0?

Also not entirely sure about the SDVX-like judgements format (Far-Lost) because a *certain* API does return Pures and Perfect Pures, but that would already require a migration anyway.

PR #1499, opened 2026-05-19
feat: add M+ support, seeds up to 2025121000

InvoxiPlayGames:
This adds support for the M+ mod for MÚSECA as well as updates and corrects previous seed data for 1+1/2 using the music DB JSON file offered on the M+ website, with a script added for updating based on that music DB.

Songs have their level based on the level that the game uses internally to calculate your curator skill, not the one displayed in-game; M+'s additional level 16s are marked as level 15s for this reason. (Maybe I should add these to the M+ Lv. 16 folder?)

---

Before this is merged I'd also like to propose the removal of the 1+1/2 Rev. B "version" for MÚSECA - the only difference is 3 songs being removed from the DB (these songs were missing in the previous Tachi seeds anyway, making it incorrect) and 3 new songs being added that are also present in 1+1/2's game files (again, 2 of these were missing in Tachi's previous seeds), all of these songs often being enabled by servers across both specs.

### DIFF
--- a/db/seeds/charts-museca.json
+++ b/db/seeds/charts-museca.json
@@ -10,7 +10,7 @@
 		"level": "3",
 		"levelNum": 3,
 		"songID": "S19d35e0b424a871d843",
-		"versions": ["1.5", "1.5-b"]
+		"versions": ["1.5", "1.5-b", "m+"]
 	},
 	{
 		"data": {
@@ -23,7 +23,7 @@
 		"level": "13",
 		"levelNum": 13,
 		"songID": "S19d35e0b424a871d843",
-		"versions": ["1.5", "1.5-b"]
+		"versions": ["1.5", "1.5-b", "m+"]
 	},
 	{
 		"data": {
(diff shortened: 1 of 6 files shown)


zkldi:
The `-b` folder made sense 5 years ago. There were servers that did not enable these songs. I have no idea what the state of affairs is on museca nowadays, but back then, people on FLO could not play everlasting message etc., it just didn't exist and we wouldn't've been able to complete folders.

Interesting catch on the broken `displayVersion`s. I'm surprised that hasn't been caught before.


*****

Prior art for tachi is to mark level 16s as level 16s, regardless of what the game visually says (c.f. redshift, V, etc.) which use `level: "16"` and `levelNum: 16`. Given that this PR doesn't add any 16s, I imagine you've got this wrong somehow?

zkldi (changes requested):
changes reqed on discord

zkldi:
+bot diff

PR #1561, opened 2026-05-23
Add new color classes for maimai DX and CHUNITHM

beer-psi:
The new color classes for maimai DX were introduced in CiRCLE PLUS, but I figure it wouldn't hurt to add it right now. Class names were also changed to Japanese to be in line with CHUNITHM and ONGEKI.

image

Added missing Platinum II and III classes for CHUNITHM.


db/migrations/20260523071801_Colour_class_updates_for_maimai_DX_and_CHUNITHM.sql
```diff
@@ -0,0 +1,26 @@
...
+WHERE game = 'maimaidx';
+
+UPDATE game_profile
+SET classes['colour'] = CASE
+    WHEN (ratings->>'naiveRating')::numeric >= 16.00 THEN classes['colour']
+    WHEN (ratings->>'naiveRating')::numeric >= 15.75 THEN '"PLATINUM_III"'
+    WHEN (ratings->>'naiveRating')::numeric >= 15.50 THEN '"PLATINUM_II"'
+    ELSE classes['colour']
+END
+WHERE game = 'chunithm';
```
zkldi: I don't think this is the right approach, you should instead update the game calculator and then you should do `insert into game_profile_dirty` for all chunithm profiles.

zkldi (changes requested):
It's weird to have a migration doing the logic and the game impl too. I think you can just update the game impl and then insert game_profile_dirty for all chunithm profiles

zkldi:
It's not that this won't work, it's just that these are pure derivations of state, and so we don't actually need to update state in the db really, just kick off a re-derivation.

beer-psi:
Changed it to just mark all CHUNITHM/maimai DX profiles as dirty. The server implementations have already been updated to use the new classes.

zkldi (approved):

PR #1589, opened 2026-05-25
feat: better arcaea judgement cell + ptt calculation fixes

nyairobi:
### Arcaea judgement cell
For pure memory (i.e. all perfect) scores, the judgement cell now displays `MAX-{n}` instead of `0-0`. MAX- is a metric hardcore Arcaea players care about, and it's only applicable for pure memory scores due to how scores are calculated.

### rg-stats Potential fix
A lot of evidence points at Arcaea not flooring its play rating values, and instead storing them untruncated. The official api returns play rating with arbitrary precision, and spreadsheet calculations show that it's impossible for the current floor-based algorithm to be correct. Because two-digit precision isn't enough to convey an arbitrary precision float, I added a tooltip with 6 digits. A migration was added to rederive all 5 Arcaea scores we currently have on the site.

### Profile rating minimumRelevantValue RED
Red is the lowest rank people remotely care about.


zkldi (approved):
Every day I regret implementing this side of things with floats and not fixed point integers like a man

PR #1591, opened 2026-05-26
feat: folder goal exemptions for metrics with chartDependentMax

nyairobi:
Really dumb hack, but in my defense, the hard split between `impl.chartSpecificValidators` and `conf.validate` based on `chartDependentMax` is also really odd.

Arcaea has a well regarded scoring system where GREAT=100% and PGREAT=100%+1. And 100% is 10 Million. This means that if a chart has 1000 notes, a perfect score can be between 10,000,000 and 10,001,000. Therefore, `score` is `chartDependentMax`, because it is–and it wouldn't make sense to set a folder goal for >10,001,000 for the same reason it wouldn't make sense to have a goal of 3000 in IIDX. But non-perfect scores behave more like in MÚSECA because the difference between CRITICAL and S-CRITICAL (on a hypothetical 120Hz MÚSECA cab) becomes insignificant, and there is definitely an incentive to set folder goals such as 9,700,000 especially with Arcaea only having 2 relevant score grades (9.8M and 9.9M).

### DIFF
--- a/typescript/common/src/config/config.ts
+++ b/typescript/common/src/config/config.ts
@@ -355,7 +355,7 @@ export function ValidateMetric(gameConfig: GameConfig, metricName: string, metri
 		return "Cannot validate a graph or nullable graph metric.";
 	}
 
-	if (conf.chartDependentMax) {
+	if (conf.chartDependentMax === true) {
 		return `This metric is chart dependent and not appropriate to check in this context.`;
 	}
 
--- a/typescript/common/src/config/game-support/arcaea.ts
+++ b/typescript/common/src/config/game-support/arcaea.ts
@@ -1,3 +1,4 @@
+import { p } from "prudence";
 import { z } from "zod";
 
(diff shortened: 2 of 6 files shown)


nyairobi:
Screenshot_20260526_113608
Screenshot_20260526_113624

zkldi (changes requested):
I think this is conflated.

I think `chartDependentMax` should stay as a boolean - it's a boolean condition after all. "Conditionally chart-dependent-max" is not the thing you're implementing here, what you're implementing is "conditionally allow folder goals to be made on this metric, anyway", and a new function should be added onto things with `chartDependentMax` with a different name.

You could add a function `allowFolderGoalsIf: (v: number) => boolean`, and I think that clears up what this is for instead of conflating it with the existing thing of chartDependentMax.

Ditto, you can implement this for `maimai`, `allowFolderGoalsIf: x => x <= 100.00`.

zkldi (approved):
lgtm

PR #1596, opened 2026-05-26
fix: arcaea notecount check

nyairobi:
Adds a check that forbids too many judgements, i.e. the sum of all judgements being greater than notecount.


zkldi (approved):
i can't wait to find out for what exciting reason this will bite you in the ass

PR #1605, opened 2026-05-28
feat: generalize ImportSQLiteForm for plaintext input + yurisaki import

nyairobi:
ImportSQLiteForm is really nice. I slightly tweaked it to allow plaintext, as I think that's more elegant than adding another dedicated importer type into the server side. I added an alternative score import method for Arcaea for players without a rooted device. Good luck extinguishing the fire in the Library of Alexandria.


zkldi (changes requested):
no sorry i think this should just be a new import type - we have lots of precedent for csv importers on the backend.

the sqlite stuff is only done on the frontend because it's grossly unsafe to let users provide sqlite files. Agreed that this is nice to do on the frontend but the backend is the right place to put this for consistency.

nyairobi:
saw how much boilerplate that involves and icba sorry

PR #1626, opened 2026-06-03
fix: update rating state on alg change in PBTable

whiteou7:
Changes in "alg" state caused by AlgSelector should be in sync with "rating" state used by ScoreCoreCells to avoid data mismatch

Example: VF7 alg selected but rating cell is displayed as VF6
image


zkldi:
in the picture you've shown me nothing seems aligned at all in terms of headers to data, what's going on here?

whiteou7:
Yeah that header issue is another problem by itself. I just wanted to point out that while vf7 is being selected, rating cells are displayed as vf6 instead of vf7, since they depend on "rating" state which is not in sync with "alg" state.

whiteou7:
Notice how the ratings are sorted wrongly (0.412 -> 0.415 -> 0.403 etc.)

PR #1633, opened 2026-06-03
Reconcile missing chart leaderboard rows

### LINKED ISSUE #1630: session "deadlock" in recalc_sessions when loading datasets without chart_leaderboards
Image

Clearly, some large amount of load causes something unpleasant to happen here in drain_stats_queues.

I will likely never get around to fixing this, so I'll put a £50 bounty on a deep dive + fix for this. That's not a joke. I am offering £50 for a good fix on this. If you send me LLM bullshit for this I will kill you. Thanks.

# I WILL NOT ACCEPT ANY LLM PRS FOR THIS. YOU WILL NOT RECEIVE YOUR MONEY.
# I CAN AND WILL SCAM YOU IF YOU ARE AN LLM. YOU ARE WASTING YOUR TOKENS

VexCode24:
Fixes #1630

## Summary
- adds a chart_leaderboard reconciliation job that finds PB partitions with missing cached leaderboard rows and refreshes them with the existing refresh_chart_leaderboard_partition function
- runs the reconciliation before session/profile dirty queues drain so restored datasets cannot hide valid PBs behind chart_leaderboard inner joins
- adds a regression test that deletes a cached leaderboard row and verifies the reconciliation rebuilds rank/out_of for the PB

## Tests
- git diff --check
- git diff --cached --check

Not run: bun/vitest tests. This local environment has no usable bun/pnpm/npm on PATH and repo dependencies are not installed.

### DIFF
--- a/typescript/server/src/lib/jobs/chart-leaderboard.test.ts
+++ b/typescript/server/src/lib/jobs/chart-leaderboard.test.ts
@@ -1,0 +1,108 @@
+import { ReconcileChartLeaderboardJob } from "#lib/jobs/reconcile-chart-leaderboard";
+import DB from "#services/pg/db";
+import { seedUser } from "#test-utils/pg-fixtures";
+import { describe, expect, it } from "vitest";
+
+function makeSongId(n: number): string {
+	return `SL${n.toString(16).padStart(20, "0")}`;
+}
+
+function makeChartId(n: number): string {
+	return `CL${n.toString(16).padStart(20, "0")}`;
+}
+
+function makeLegacyId(n: number): string {
+	return `leaderboard_${n.toString(16).padStart(28, "0")}`;
+}
+
+async function insertSongAndChart(songId: string, chartId: string) {
+	await DB.insertInto("song")
+		.values({
+			id: songId,
+			legacy_id: 9_020_000,
+			game_group: "iidx",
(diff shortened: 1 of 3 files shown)


VexCode24:
A bit more detail on the failure mode this patch is targeting:

- `chart_leaderboard` is now a cached table keyed by `pb.row_id`, not the old always-derived view.
- The PB/profile read paths inner-join `chart_leaderboard`, so if a dataset restore leaves PB rows without matching cache rows, those PBs disappear from downstream recalculation instead of showing up with a nullable rank.
- The migration already has the authoritative ranking logic in `refresh_chart_leaderboard_partition(chart_id, lens)`, so this PR reuses that rather than copying the window-ordering rules into TypeScript.
- The new reconcile job only looks for missing cache rows, groups by the affected `(chart_id, lens)` partitions, and refreshes those partitions.
- The drain hooks run the repair after PB dirty processing and before session/profile dirty processing, so restored datasets can settle before `recalc_sessions` consumes downstream queues.
- The regression test simulates the bad restore state by deleting a cached leaderboard row while leaving the PB intact, then verifies the repair restores `rank/out_of`.

I kept this intentionally narrow: it repairs the missing-cache state described in #1630 without changing the existing trigger-maintained ranking behavior for normal PB writes.

zkldi:

PR #1634, opened 2026-06-03
Fix score dirty queue deadlocks during stats drains

yZangEren:
Addresses #1630.

## Deep dive

The screenshot shows `drain_stats_queues` deadlocking inside the `enqueue_session_dirty()` trigger while PostgreSQL is inserting into `session_dirty` with `ON CONFLICT DO NOTHING`.

The deadlock comes from `drainScoreRederive()` processing multiple chart rederive jobs concurrently. Each job bulk-updates `score`, and the row-level score triggers enqueue `pb_dirty`, `session_dirty`, and `game_profile_dirty` rows as every score row is touched. When two chart transactions hit overlapping sessions/profiles in different score-row orders, `ON CONFLICT DO NOTHING` still waits on uncommitted unique-index entries, so the transactions can wait on each other.

## Fix

- Replace the score dirty-queue row-level triggers with statement-level transition-table triggers.
- Enqueue distinct dirty keys once per UPDATE/INSERT/DELETE statement and sort those keys before insertion, so overlapping transactions take unique-index locks in a deterministic order.
- Cover `pb_dirty`, `session_dirty`, and `game_profile_dirty`, including UPDATE cases where a score changes session or PB/profile identity.
- Keep the stats-queue cron bounded by draining each queue once per tick budget instead of repeatedly resetting per-queue caps inside an outer loop.

## Tests added

- Dirty-key dedupe for batched score inserts.
- Session move behavior marks both old and new sessions dirty.
- Tick-budget helper drains until empty or until the cap is reached.

## Verification

- `git diff --check` passed locally before commit.
- I could not run the Bun test suite in this Windows environment because Bun is not installed here and the installer download timed out. The patch is intentionally narrow and includes focused tests for the changed behavior.

### DIFF
--- a/db/migrations/20260604032000_score_dirty_statement_triggers.sql
+++ b/db/migrations/20260604032000_score_dirty_statement_triggers.sql
@@ -1,0 +1,213 @@
+-- The score dirty-queue triggers used to run once per row. During large
+-- rederive/import UPDATEs, parallel transactions could insert overlapping
+-- session_dirty / pb_dirty / game_profile_dirty keys in different row orders
+-- and deadlock inside the unique indexes despite ON CONFLICT DO NOTHING.
+--
+-- Use AFTER STATEMENT transition tables instead: each statement inserts each
+-- dirty key once, in deterministic key order.
+
+DROP TRIGGER "score_pb_dirty" ON score;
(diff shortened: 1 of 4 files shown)


zkldi:

PR #1643, opened 2026-06-06
Update last seen after the request is finished

### LINKED ISSUE #1641: Changing username raises an internal server error

beer-psi:
Closes #1641

Shot in the dark since I can't reproduce this issue on local dev. My best guess at why this is happening is, in order:
- The promise updating `last_seen` yields before executing the query
- `ACTION_ChangeUsername` begins a serializable transaction
- The promise updating `last_seen` executes the query -> account row is updated mid-transaction (hence `could not serialize access due to concurrent update`)

I confirmed that this is most likely the issue by enabling invisible mode through the `PATCH /api/v1/users/:userID/settings` API and then attempting to change my username, in which case the change goes through correctly.

----

Also returns `nextAvailableChange` instead of `nextChange` in `GET /users/:userID/last-username-change` since that is what the original key was when username change was added.

https://github.com/zkldi/Tachi/blob/8f4955812d8fd05eba1a9892558db8fe40d2fc48/client/src/app/pages/dashboard/users/UserSettingsPage.tsx#L305-L308

### DIFF
--- a/typescript/client/src/app/pages/dashboard/users/UserSettingsPage.tsx
+++ b/typescript/client/src/app/pages/dashboard/users/UserSettingsPage.tsx
@@ -306,7 +306,7 @@ function ChangePasswordForm({ reqUser }: { reqUser: UserDocument }) {
 
 function ChangeUsernameForm({ reqUser }: { reqUser: UserDocument }) {
 	const lastUsernameChange = useApiQuery<
-		{ canChange: false; nextAvailableChange: integer | null } | { canChange: true }
+		{ canChange: false; nextChange: integer | null } | { canChange: true }
 	>(`/users/${reqUser.id}/last-username-change`);
 
 	const nameChangeFormik = useFormik({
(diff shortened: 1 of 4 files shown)


zkldi:
> Also returns nextAvailableChange instead of nextChange in GET /users/:userID/last-username-change since that is what the original key was when username change was added.


Change this back imo, no need to add this in as a change

beer-psi:
> > Also returns nextAvailableChange instead of nextChange in GET /users/:userID/last-username-change since that is what the original key was when username change was added.
> 
> Change this back imo, no need to add this in as a change

I changed it to fix frontend since currently that just shows "you last changed your username on 1/1/1970" due to the property mismatch. Should I break it out into a separate PR?

zkldi:
change the frontend and the types imo. I've messed up a bit here with the action values, but it's generally best not to change action inputs/outputs if possible - they're stored in the `"action"` table as an audit log

zkldi (approved):

PR #1658, opened 2026-06-10
seeds: update ONGEKI to June 4th, 2026

AlexC233:
Song info from https://ongeki.sega.jp/assets/json/music/music.json
Chart info from https://wikiwiki.jp/gameongeki/

### DIFF
--- a/db/seeds/charts-ongeki.json
+++ b/db/seeds/charts-ongeki.json
@@ -6304,7 +6304,7 @@
 	},
 	{
 		"data": {
-			"chartViewURL": "https://sdvx.in/ongeki/luna/01038luna2.htm",
+			"chartViewURL": "https://sdvx.in/ongeki/luna/01038luna.htm",
 			"displayVersion": "オンゲキ bright MEMORY Act.2",
 			"inGameID": 8003,
 			"isBonusTrack": false,
@@ -6327,6 +6327,7 @@
 	},
 	{
 		"data": {
+			"chartViewURL": "https://sdvx.in/ongeki/01/01038mst.htm",
 			"displayVersion": "オンゲキ",
 			"inGameID": 36,
 			"isBonusTrack": false,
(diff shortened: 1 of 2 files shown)


nyairobi:
Who are you
wikiwiki is not a reliable source

AlexC233:
> Who are you wikiwiki is not a reliable source

Oh I'm sorry, I didn't know. I just assumed so since the silentblue wiki said they took it from there. What source should I be using?

nyairobi:
Thanks for the contribution but you could have communicated it with me ahead of time

Seeding instructions are at https://github.com/zkldi/Tachi/blob/main/typescript/seeds-scripts/rerunners/ongeki/README.md

wikiwiki is known to lie about notecounts or omit them (especially for BAS and ADV) and incorrect notecount data (maxPlatinumScore) will mess with p-stars calculation. And this is on top of internal levels also being semi-madeup. If you can vouch for maxPlatinumScore being correct, this is possibly mergable (that's ultimately up to zk); I'll take a closer look when I get home

AlexC233:
I see. Thanks for letting me know about all this. I will verify the maxPlatinumScore with the values on ongeki-net then.

AlexC233:
All the maxPlatScore values now match what is on ongeki-net.

db/seeds/charts-ongeki.json
```diff
@@ -6304,7 +6304,7 @@
 	},
 	{
 		"data": {
-			"chartViewURL": "https://sdvx.in/ongeki/luna/01038luna2.htm",
+			"chartViewURL": "https://sdvx.in/ongeki/luna/01038luna.htm",
```
nyairobi: Where does this change come from? This is the LUN 0 chart.

db/seeds/charts-ongeki.json
```diff
@@ -101921,5 +101922,1745 @@
...
+		"levelNum": 11.7,
+		"songID": "S19eb183204c24ceec92",
+		"versions": [
+			"refresh",
+			"refreshOmni"
+		]
+	},
+	{
+		"data": {
+			"chartViewURL": "",
```
zkldi: the empty strings here are wrong, it should be unset if not known

zkldi (changes requested):
Can't vouch for the accuracy of this data (please avoid using wiki's, they are _frequently_ wrong) but the empty str stuff here is definitely wrong and needs fix.

AlexC233:
Ok I’ll fix it on Tuesday when I get back to my computer.
I’ll get the internal levels for these charts from https://x.com/ongeki_level and the maxPlatScore from ongeki-net.

nyairobi (changes requested):

db/seeds/charts-ongeki.json
AlexC233: Yeah that’s odd, it shouldn’t have changed. scrape-sdvx-in.ts swapped it and I did not notice.

nyairobi: right, I haven't adjusted scrape-sdvx-in for the perfect shining merger; my bad

AlexC233:
I've verified the internal levels for these charts from https://x.com/ongeki_level and the maxPlatScore from ongeki-net.

nyairobi:
時空を超えて久しぶり's genre should be `LUNATIC` not `VARIETY`
The 4 bonus tracks' genre should be `ボーナストラック` not `オンゲキ` 

(This makes me realize `isBonusTrack` is technically no longer necessary in Tachi3)

nyairobi:
lgtm

nyairobi (approved):

zkldi (approved):

PR #1675, opened 2026-06-16
Set a default of 0 when a rating value is null when calculating rank

### LINKED ISSUE #1674: profile rankings say you're #1 if you have a "null" rating on the metric
Obvious example is BPI: players with no bpi-eligible scores ("No Data.") say BPI #1 on their profiles

HutchyBen:
ficxs #1674

### DIFF
--- a/typescript/server/src/utils/user.test.ts
+++ b/typescript/server/src/utils/user.test.ts
@@ -131,15 +131,17 @@ describe("GetUsersRankingAndOutOf", () => {
 		expect(result).toEqual({ ranking: 1, outOf: 2 });
 	});
 
-	it("returns ranking 1 when user rating is null (no one can be strictly greater than null)", async () => {
+	it("returns worst ranking when user rating is null", async () => {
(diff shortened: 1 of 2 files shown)


zkldi (changes requested):
not correct! bpi can be negative. null should sort lower than _all_ numbers, not just 0. Don't solve this with `?? -Infinity` either. There is a proper way to do this with sql.

zkldi (approved):

HutchyBen:
unapprove!! unapprove ill see what that fail was

HutchyBen:
the test is literally LOL?
	it("returns ranking 1 when user rating is null (no one can be strictly greater than null)", async () => {

HutchyBen:
ill change the test add like another player and say null user should be last

zkldi:
lol yeah that test is llm nonsense, thanks

HutchyBen:
> lol yeah that test is llm nonsense, thanks

classic

PR #1676, opened 2026-06-16
Run sort seeds to make just fmt-check be quiet

HutchyBen:
why cant like committing just like auto format instead of whining that its broken?


zkldi:
fixed in #1678

PR #1705, opened 2026-06-29
fix: dynamic baseline for ongeki score graphs

nyairobi:
Vastly improves clarity for most graphs people are likely to care about.

Before
Screenshot_20260630_005759

After
Screenshot_20260630_005736


nyairobi:
the checks are failing because of #1704

PR #1729, opened 2026-08-13
seeds: WACCA 4.06 update

szymex73:
Couple of new songs and three re-rates

The WACCA re-runner needs some changes to work, but I believe those changes will come in a separate PR from GRIM657


szymex73:
I've spotted a mistake through the diff view so don't merge yet, rerunner doesn't insert unique IDs it seems

szymex73:
I think it's correct now, or at least it appears correct in the seed diff

szymex73:
I've also included a seeds folder fix for WACCA's plus grouping since 14+ charts were not being included in the 14-15 folder (not doing the same for Reverse as it did not have any 14+ charts and that group of songs is not going to receive one).
Was reported [here](https://discord.com/channels/683092550160351261/683093664192266282/1541790526464860240)

PR #1736, opened 2026-08-30
feat: arcaea 7.0

nyairobi:
They added an alias for their max difficulty (akin to SDVX), so I'm creating something modeled on `sdvxInGameID` for future-proofing.

The rating (potential) algorithm has changed from B30+R10 to B50+B10, which is reasonable enough to be added to the site as-is. This algorithm stays under `naivePotential` as to not destroy existing graphs, but it's renamed to `Potential` in the client.

This is a breaking change (`inGameStrID` is removed as a valid match type) because the migration of `Beyond` scores to `AnyBeyond` is one-time.


nyairobi:
Somewhat tested with a few custom imports and with anon-kamai.

PR #1737, opened 2026-09-01
Add missing song lengths and inGameIDs for Ongeki Refrech Act 1 seeds

Kumubou:
This is an update to the seeds for Ongeki that does three things:

- Adds song lengths for the Refresh Act 1 songs missing them
- Adds inGameIDs for the Refresh Act 1 charts missing them
- Adds a bunch of additional sdvx.in links for charts

Apologies if I botched something here (as it's my first time trying to update seeds), but it does pass all tests and load in my local dev instance and I did cross reference this against both the sdvx.in data and data generated from [other sources], so it should check out. (...hopefully?)

### DIFF
--- a/db/seeds/charts-ongeki.json
+++ b/db/seeds/charts-ongeki.json
@@ -5568,7 +5568,6 @@
 	},
 	{
 		"data": {
-			"chartViewURL": "https://sdvx.in/ongeki/01/01055exp.htm",
 			"displayVersion": "オンゲキ",
 			"inGameID": 78,
 			"isBonusTrack": false,
@@ -6975,7 +6974,6 @@
 	},
 	{
 		"data": {
-			"chartViewURL": "https://sdvx.in/ongeki/01/01044exp.htm",
 			"displayVersion": "オンゲキ",
 			"inGameID": 68,
 			"isBonusTrack": false,
@@ -7290,7 +7288,6 @@
 	},
 	{
 		"data": {
-			"chartViewURL": "https://sdvx.in/ongeki/01/01047exp.htm",
 			"displayVersion": "オンゲキ",
 			"inGameID": 71,
 			"isBonusTrack": false,
@@ -7427,7 +7424,7 @@
 		"data": {
 			"chartViewURL": "https://sdvx.in/ongeki/luna/01048luna.htm",
 			"displayVersion": "オンゲキ Re:Fresh",
-			"inGameID": 8189,
+			"inGameID": null,
 			"isBonusTrack": false,
 			"maxPlatScore": 2948
 		},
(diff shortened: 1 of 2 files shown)


nyairobi:
Hello, thank you for the PR.
Have the `inGameID`s been added only by using the in-repo scripts and the song data export mod?

Kumubou:
> Hello, thank you for the PR. Have the `inGameID`s been added only by using the in-repo scripts and the song data export mod?

The inGameIDs were generated using `parse-music-data-output.json` currently in the repo, but it did take a bit of massaging to get the inGameID update to actually take. The script wasn't actually updating the inGameID field, so I had to remove the affected chart entries from my local instance and then run the script again (and then run another jq command so that the chart IDs stayed the same, which the diff here verifies). The current version of this script also tried to revert the Re:Master change, so I had to throw those specific changes out. I did also create a small Bash script to compare the added inGameIDs between the data in the generated `music.json` and the output in the `charts-ongeki.json` and `songs-ongeki.json` seed files to verify correctness.

I couldn't get the `parse-song-duration.ts` script to work _at all_, so I ended up coming up with a short Bash script to match the songIDs here to their inGameIDs and then used that to find the correct audio file and ran `vgmstream-cli` against that. So I'm a bit less confident that this data is 100% -- I went ahead and attached the shell script I used to add these, if you want to review that. [parse_ongeki_song_lengths.sh](https://github.com/user-attachments/files/31760515/parse_ongeki_song_lengths.sh)

songSeedsLocation and chartSeedsLocation should be straightforward, just set the full path to the song and chart seeds file you're working with there. awbDirectory should be the path for a directory that has _all_ of the AWB files you're looking to read merged together into one directory, which you can pull from the game data (yes, this is just me being lazy and not wanting to sort out how to navigate all of the game data on the fly, lol).

The script is also expecting a `ongeki_song_ids` file in the same location as the script, which is a text file that has the inGameIDs of the songs you are looking to update, one ID per line. One obvious improvement here would be to walk through the songs-ongeki.json seeds to find which songs have a null length and then use that generated list, but that would then require either working out how to navigate all of the game data or just shotgun all of the game's audio into that directory (which is inelegant, but would work).

`scrape-sdvx-in.ts` threw a bunch of warnings but it seemed to then add a bunch of additional links without issue, so it seems like that worked as expected. It did mangle the entry for Perfect Shining!! Lunatic, but I'm guess that happens because it's the only song in the game with both a Re:Master and a Lunatic chart, so I manually reverted that.

nyairobi:
It seems like at least [some of the fixes](https://github.com/zkldi/Tachi/pull/1427/changes#diff-919974ee53c9021a6f04fb8603743603633a7e1efa193d43b180eb97c1ccda67) to the scripts were quietly reverted with the switch to v3.

Kumubou:
That older script (I did search the main repo and some forward forks... didn't think to check previous versions though) still has the issue where it's not updating inGameIDs for existing chart entries (it seems like that function only current updates difficulty ratings and version details), so the hack job here would still need to be done.

nyairobi:
No hackjobs. We'll fix the scripts.

nyairobi:
`inGameID`s and `duration`s match the data generated with #1739 using the same music.json, so this is good to go.

PR #1740, opened 2026-09-03
seeds: WACCA 4.06 update pt. 2

szymex73:
Couple of new songs and new inf diffs for existing ones


szymex73:
I see the ci fail, same bug as before snuck into the rerunner when I re-did the changes since I didn't save the copy from last time. I'll fix the seeds and send a PR with my rerunner update to avoid this in the future

PR #1744, opened 2026-09-12
fix: port basic seeds scripts to v3

nyairobi:
Ported the top-level seeds scripts to the v3 seeds format.


nyairobi:
Everything is failing because seeds are unformatted on main

